PR #59, opened 2022-09-23
Liquidations

DaigaroCota:


brozorec:
@DaigaroCota can you add a section in the general [doc](https://www.notion.so/Fuji-V2-Himalaya-70bdd796b355435b8ef958b136cc44fb#8fdb40485cf5445ab1b940aa9c8c136f) on how liquidation works. Also, maybe some more comments in the contract about it may be helpful.

brozorec:
@DaigaroCota @ben132333 how do we use [liqRatio](https://github.com/Fujicracy/fuji-v2/blob/00d5142751f2e698487842527d3e878ae1088e9e/packages/protocol/src/vaults/borrowing/BorrowingVault.sol#L69) when performing a liquidation?

ben132333:
The liqRatio variable is not used when performing a liquidation. I was unaware of it. The healthFactor determines + determineLiquidatorFactor determine when a user can be liquidated. Should liqRatio be integrated?

DaigaroCota:
> The liqRatio variable is not used when performing a liquidation. I was unaware of it. The healthFactor determines + determineLiquidatorFactor determine when a user can be liquidated. Should liqRatio be integrated?

This was a miss on my side, but it is an easy fix, and I will take care of it. Essentially the maxLTV value needs to be replaced by liqRatio. This is to avoid users that take a debt position for the first time, basically exposing themselves to liquidation but borrowing the maximum amount. That is why liqRatio should be > maxLTV. Allowing at least a window between having maxed out borrowing amount and liquidation possibility.

brozorec:
> > 
> 
> I was thinking we do not over-complicate the function in the BorrowingVault, so yes the intention was there to just keep it simple. For me we can either create a separate contract or we include in the router. Though, my only comment would be that we need to check how big in bytecode with optimization runs is getting. Perhaps also in terms of access to the liquidation function it becomes cleaner on a seperate contract? That way we can include helper functions that can determine if a user is insolvent (profit-less liquidations) before execution of an array of users.

I agree with this. However, I suggest we can prepare the ground for it here. If we want this function to be called by another contract (the router or another one), we need at least one new input param to identify the receiver, the same way `deposit` and `payback` work:
```
liquidate(address owner, address receiver) {
  ...
  address _receiver = receiver != address(0) ? receiver : caller;
  mint(_receiver, shares);
}
```

DaigaroCota:
> > > 
> > 
> > 
> > I was thinking we do not over-complicate the function in the BorrowingVault, so yes the intention was there to just keep it simple. For me we can either create a separate contract or we include in the router. Though, my only comment would be that we need to check how big in bytecode with optimization runs is getting. Perhaps also in terms of access to the liquidation function it becomes cleaner on a seperate contract? That way we can include helper functions that can determine if a user is insolvent (profit-less liquidations) before execution of an array of users.
> 
> I agree with this. However, I suggest we can prepare the ground for it here. If we want this function to be called by another contract (the router or another one), we need at least one new input param to identify the receiver, the same way `deposit` and `payback` work:
> 
> ```
> liquidate(address owner, address receiver) {
>   ...
>   address _receiver = receiver != address(0) ? receiver : caller;
>   mint(_receiver, shares);
> }
> ```

Implemented as suggested, only difference is that I forced receiver to never be address zero. If you are liquidating, either you pass yourself (address) as receiver or someone else (even if "you" is a contract).

brozorec (approved):

PR #70, opened 2022-10-05
Streams

brozorec:
Add methods that return `Observable`s on which we can subsribe.


packages/sdk/src/entities/BorrowingVault.ts
```diff
@@ -157,10 +180,12 @@ export class BorrowingVault {
   /**
    * Creates a connection by setting an rpc provider.
    *
-   * @param configParams - {@link ConfigParams} object with infura and alchemy ids
+   * @param configParams - {@link ChainConfigParams} object with infura and alchemy ids
    */
-  setConnection(configParams: ConfigParams): BorrowingVault {
-    this.rpcProvider = Config.rpcProviderFrom(configParams, this.chainId);
+  setConnection(configParams: ChainConfigParams): BorrowingVault {
```
doliG, line 185: Maybe `ChainConfigParam` could be shortened with `ChainConfig` (almost the same meaning 🤔 )

packages/sdk/src/entities/BorrowingVault.ts
```diff
@@ -275,6 +298,30 @@ export class BorrowingVault {
...
 
+  /**
+   * Returns a stream of deposit and borrow balances for an account.
+   *
+   * @param account - user address, wrapped in {@link Address}
+   * @throws if {@link setConnection} was not called beforehand
+   */
+  getBalancesStream(account: Address): Observable<AccountBalances> {
+    invariant(this.contract && this.wssProvider, 'Connection not set!');
+    const filters = [
```
doliG, line 309: We don't need `flashclose | liquidate` in our filters ?

packages/sdk/src/entities/ChainConnection.ts
```diff
@@ -0,0 +1,60 @@
...
+import { ChainId } from '../enums';
+import { ChainConfigParams } from '../types';
+
+type ChainConnectionParams = {
+  rpcProvider: StaticJsonRpcProvider;
+  wssProvider?: WebSocketProvider;
+};
+
+export class ChainConnection {
+  private static _config: {
```
doliG, line 21: can be converted to Record 🤔

packages/sdk/test/entities/BorrowingVault.test.ts
```diff
@@ -8,14 +8,17 @@ import { Address, BorrowingVault } from '../../src/entities';
...
+  ChainConfigParams,
   DepositParams,
   PermitParams,
   RouterActionParams,
 } from '../../src/types';
 
-const config: ConfigParams = {
+const config: ChainConfigParams = {
   infuraId: process.env.INFURA_ID ?? '',
+  alchemy: {
```
doliG, line 19: alchemy should be also passed to sdk when constructing the classe ?

doliG (approved):

packages/sdk/src/entities/ChainConnection.ts
brozorec: `[chainId in ChainId]?` allows optionality while it's not the case with Record afaik

packages/sdk/src/entities/BorrowingVault.ts
brozorec: good catch about liquidate

packages/sdk/test/entities/BorrowingVault.test.ts
brozorec: done

packages/sdk/src/entities/BorrowingVault.ts
brozorec: done

PR #92, opened 2022-10-17
Vault fuzz tests

pedrovalido:
Add fuzz arguments to all test_ functions where they are missing


packages/protocol/test/Vault.t.sol
```diff
@@ -30,6 +37,10 @@ contract VaultTest is DSTestPlus {
   uint256 bobPkey = 0xB;
   address bob = vm.addr(bobPkey);
 
+  // These test prices should be inverse of each other.
+  uint256 public constant TEST_USD_PER_ETH_PRICE = 2000e18;
+  uint256 public constant TEST_ETH_PER_USD_PRICE = 5e14;
```
brozorec, line 42: it's a good idea to set these values as constants but why prefixing them with `TEST_`?

packages/protocol/test/Vault.t.sol
```diff
@@ -103,6 +116,97 @@ contract VaultTest is DSTestPlus {
...
+  {
+    uint8 debtDecimals = 18;
+    uint8 assetDecimals = 18;
+    uint256 liqRatio = 80 * 1e16;
+
+    uint256 price = oracle.getPriceOf(address(debtAsset), address(asset), debtDecimals);
+    uint256 hf = SafeMath.div(
+      SafeMath.mul(amount * liqRatio, (price - priceDrop)),
+      SafeMath.mul(borrowAmount * 1e18, 10 ** assetDecimals)
+    );
```
brozorec, line 198: do we actually need `SafeMath`?

packages/protocol/test/Vault.t.sol
```diff
@@ -103,6 +116,97 @@ contract VaultTest is DSTestPlus {
...
+
+    return hf > 95 && hf < 100;
+  }
+
+  function _utils_add(uint256 a, uint256 b) internal pure returns (uint256) {
+    uint256 c = a + b;
+    require(c >= a);
+    require(c >= b);
```
brozorec, line 206: can be shorten to `require(c >= a && c >= b)`

packages/protocol/test/Vault.t.sol
```diff
@@ -281,40 +411,65 @@ contract VaultTest is DSTestPlus {
...
+    vm.assume(
+      priceDrop
+        > _utils_getLiquidationThresholdValue(
+          address(vault), TEST_USD_PER_ETH_PRICE, amount, borrowAmount
+        )
+    );
+
+    uint256 price = oracle.getPriceOf(address(debtAsset), address(asset), 18);
+    uint256 liqRatio = IBVaultTestingExtOnly(address(vault)).liqRatio();
+    uint256 priceDropThresholdToMaxLiq = price - ((95 * borrowAmount * 1e34) / (amount * liqRatio));
```
brozorec, line 427: it's not clear to me how we got to `1e34`, maybe assign this value to a const whose name hints at its meaning or its derivation

brozorec (changes requested):

packages/protocol/test/Vault.t.sol
pedrovalido: 1e34 is because the 95 is actually 95e16 and borrowAmount is 1e18. Changed it to: 95e16 * borrowAmount * 1e18

packages/protocol/test/Vault.t.sol
```diff
@@ -11,8 +11,15 @@ import {IVault} from "../src/interfaces/IVault.sol";
 import {ILendingProvider} from "../src/interfaces/ILendingProvider.sol";
 import {BorrowingVault} from "../src/vaults/borrowing/BorrowingVault.sol";
 import {BaseVault} from "../src/abstracts/BaseVault.sol";
+import {SafeMath} from "openzeppelin-contracts/contracts/utils/math/SafeMath.sol";
+
+interface IBVaultTestingExtOnly is IVault {
+  function liqRatio() external view returns (uint256);
+}
```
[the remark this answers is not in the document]
pedrovalido: I think we don't actually need the liquidation ratio from the vault at all as the purpose is to test it as well. So, removed the vault.liqRatio() and replaced it with the constant LIQUIDATION_RATIO created from the above suggestion

pedrovalido:
@brozorec let me know if all changes have been implemented :)

DaigaroCota (approved):
These updates look good to me. I would only say that perhaps on a new issue, we handle the fuzzing including zero., and removing the `vm.assume(value > 0)`. I mention this because our function implementations should have error cases to handle zero inputs. Therefore the test should including a logic in where if fuzzing sends zero, this error is expected otherwise check something else as we already have.
I think it should not affect this merge, I just propose we do it in a new issue.

brozorec (approved):

PR #93, opened 2022-10-17
Token prices and balances

brozorec:


packages/sdk/src/Sdk.ts
```diff
@@ -69,16 +68,44 @@ export class Sdk {
...
 
+  /**
+   * Retruns the token balances of an address in a batch.
+   * Throws an error if `chainId` is different from each `token.chainId`.
+   *
+   * @param tokens - array of {@link Token} from the same chain
+   * @param account - user address, wrapped in {@link Address}
+   * @param chainId - ID of the chain
+   */
+  getBatchTokenBalances(
```
doliG, line 86:
```suggestion
  getBalancesFor(
```

Suggestion to be consistent with `getBalanceFor`

doliG (approved):
I love how fast `tsup` is compared to `tsdx`, thanks 🙏 

Maybe you forget to remove tsdx in `package.json` ?

packages/sdk/src/Sdk.ts
brozorec: You're right about pointing out the consistency and tbh my first version of this method was like yours. However, there's a slight difference between this one and the others: this accepts `Token`s and the other `Currency`. If you think this difference doesn't justify a change in the naming suite, I'll be happy to go with `getBalancesFor`.

brozorec:
> Maybe you forget to remove tsdx in `package.json` ?

I'm still using it to run the tests. Yet, I have to learn how to configure and run jest.

packages/sdk/src/Sdk.ts
doliG: In that case maybe getTokensBalanceFor() ? My main concern was about the `batch` in the name, because it's redundant with the typing

brozorec: let's go then with `getTokenBalancesFor()`

PR #94, opened 2022-10-17
Access control

DaigaroCota:


packages/protocol/src/helpers/SystemAccessControl.sol
```diff
@@ -0,0 +1,37 @@
...
+ * @title SystemAccessControl.
+ * @author Fujidao Labs
+ * @notice Helper contract that should be inherited by contract implementations that
+ * should call {Chief} for access control checks.
+ */
+
+import {IChiefAccessHelper} from "../interfaces/IChiefAccessHelper.sol";
+import {CoreRoles} from "./CoreRoles.sol";
+
+contract SystemAccessControl is CoreRoles {
```
brozorec, line 14: Should it be an abstract contract? 
I also suggest placing this contract and all the others in a new directory "access".

packages/protocol/src/interfaces/IChiefAccessHelper.sol
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * @title Chief helper interface.
+ * @author Fujidao Labs
+ * @notice Defines interface for {Chief} access control operations.
+ */
+
+import {IAccessControl} from "openzeppelin-contracts/contracts/access/IAccessControl.sol";
+
+interface IChiefAccessHelper is IAccessControl {
```
brozorec, line 12: Why should be called "Helper"? I guess we can call it simply "IChief"

packages/protocol/src/interfaces/IVault.sol
```diff
@@ -12,8 +12,14 @@ import {ILendingProvider} from "./ILendingProvider.sol";
 import {IFujiOracle} from "./IFujiOracle.sol";
 
 interface IVault is IERC4626 {
-  /// Events
+  struct RebalanceAction {
```
brozorec, line 15: maybe this can be handled in a dedicated PR

brozorec (commented):
@DaigaroCota very good and efficient job :clap:

packages/protocol/src/helpers/SystemAccessControl.sol
DaigaroCota: Abstract contracts are only ones that do not implement certain functions. In this case SystemAccessControl is a very simple contract (or class) that just inherits the core roles and makes available generic modifiers that will be used across different contracts. It is not required to be marked as 'abstract'.

I created the folder "access" 
access/
├── CoreRoles.sol
├── SystemAccessControl.sol
└── TimeLock.sol

packages/protocol/src/interfaces/IChiefAccessHelper.sol
DaigaroCota: renamed. Only comment here is that I am not enforcing the interface over Chief contract. Only `timelock()` is needed to be known externally by our contracts.

packages/protocol/src/interfaces/IVault.sol
DaigaroCota: I needed a place holder function to apply rebalancer role restriction. Do not intend to work on it in this PR. I added "TODO" statements.

DaigaroCota:
> btw plz review https://github.com/Fujicracy/fuji-v2#branch-and-pr-naming 😉

Capitalize A?

packages/protocol/foundry.toml
```diff
@@ -3,7 +3,7 @@ src = 'src'
 out = 'out'
 libs = ['lib']
 optimizer = true
-optimizer_runs = 10000
+optimizer_runs = 1000
```
brozorec, line 6: should we use the default 200?

packages/protocol/src/access/TimeLock.sol
```diff
@@ -0,0 +1,330 @@
...
+ * @notice Enforce a time lock in Fuji-V2 risk parameter change operations. This contract is
+ * based on OpenZeppelin-TimelockController; however, all access control calls
+ * are delegated to {Chief}.
+ * Changes are indicated through documentation
+ * `SupportsInterface` and `onReceived` token functions were removed from the OZ implementation.
+ */
+
+import {SystemAccessControl} from "./SystemAccessControl.sol";
+
+contract TimeLock is SystemAccessControl {
```
brozorec, line 16: I partially see the point of having our own implementation but why not inherit from OZ's "TimelockController" and only have the chief set in the constructor?

btw OZ use "Timelock" and not "TimeLock"

packages/protocol/src/helpers/AddrMapper.sol
```diff
@@ -1,59 +1,94 @@
...
-    _addrMapping[keyAddr] = returnedAddr;
+  function setMapping(string memory providerName, address keyAddr, address returnedAddr)
+    public
+    override
+    onlyTimeLock
+  {
+    if (!_isProviderNameAdded[providerName]) {
+      _isProviderNameAdded[providerName] = true;
+    }
+    _addrMapping[providerName][keyAddr] = returnedAddr;
```
brozorec, line 67: need to push `providerName` to `_providerNames`

packages/protocol/src/helpers/AddrMapper.sol
```diff
@@ -1,59 +1,94 @@
...
     public
     override
-    onlyOwner
+    onlyTimeLock
   {
-    _addrNestedMapping[keyAddr1][keyAddr2] = returnedAddr;
+    if (!_isProviderNameAdded[providerName]) {
+      _isProviderNameAdded[providerName] = true;
+    }
+    _addrNestedMapping[providerName][keyAddr1][keyAddr2] = returnedAddr;
```
brozorec, line 91: need to push `providerName` to `_providerNames`

packages/protocol/src/access/TimeLock.sol
brozorec: No, actually, I don't get the need to have our own implementation. On the contrary, this will prevent us from using the tools for managing timelocks like Defender https://defender.openzeppelin.com/#/admin/addTimelock

packages/protocol/foundry.toml
DaigaroCota: This is how optimization_runs works. The higher the optimization_runs the cheaper gas cost to call functions for users, but the higher the contract bytecode. The lower the optimization runs the higher the gas cost for users, but lower the contract bytecode. For the user benefit, you try to run as high optimizations as you can, as the gas cost of your contract will be paid for as long as the contract is used.

packages/protocol/src/access/TimeLock.sol
DaigaroCota: Under the assumption we wanted to handle all access control authorization at a central hierarchy in the Chief contract, the Openzeppelin timelock implements its own separate AccessControl. It defines its roles within it. 

That would mean that roles like Default_ADMIN within Chief, have no control whatsoever on the Timelock. 

I propose we first experiment with Defender timelock on a testnest and see what are the "special" features about it. Otherwise, we commit ourselves that when our team scales, we will have to rely more and more on the paid-version of it.

packages/protocol/src/helpers/AddrMapper.sol
DaigaroCota: good catch this is fixed.

packages/protocol/src/helpers/AddrMapper.sol
DaigaroCota: good catch this is fixed.

packages/protocol/src/access/TimeLock.sol
DaigaroCota: Renamed TimeLock to Timelock

packages/protocol/src/Chief.sol
```diff
@@ -1,42 +1,163 @@
...
-import {Ownable} from "openzeppelin-contracts/contracts/access/Ownable.sol";
+import {AccessControl} from "openzeppelin-contracts/contracts/access/AccessControl.sol";
+import {Address} from "openzeppelin-contracts/contracts/utils/Address.sol";
+import {IPausableVault} from "./interfaces/IPausableVault.sol";
 import {IVaultFactory} from "./interfaces/IVaultFactory.sol";
+import {AddrMapper} from "./helpers/AddrMapper.sol";
+import {CoreRoles} from "./access/CoreRoles.sol";
+import {Timelock} from "./access/Timelock.sol";
+
+import "forge-std/console.sol";
```
brozorec, line 12: remove log

brozorec (changes requested):
@DaigaroCota check how I propose to use OZ Timelock https://github.com/Fujicracy/fuji-v2/pull/100

packages/protocol/src/Chief.sol
```diff
@@ -1,42 +1,163 @@
...
 import {IVaultFactory} from "./interfaces/IVaultFactory.sol";
+import {AddrMapper} from "./helpers/AddrMapper.sol";
+import {CoreRoles} from "./access/CoreRoles.sol";
 
 /// @dev Custom Errors
-error ZeroAddress();
-error NotAllowed();
+error Chief__setTimelock_ZeroAddress();
+error Chief__deployVault_ZeroAddress();
+error Chief__deployVault_FactoryNotAllowed();
```
brozorec, line 16: still not respecting the convention

packages/protocol/src/Chief.sol
```diff
@@ -1,42 +1,163 @@
...
+
+  event OpenVaultFactory(bool state);
   event DeployVault(address indexed factory, address indexed vault, bytes deployData);
   event AddToAllowed(address indexed factory);
   event RemoveFromAllowed(address indexed factory);
 
-  mapping(address => bool) public vaults;
+  address public timelock;
+  address public addrMapper;
+  bool public openVaultFactory;
```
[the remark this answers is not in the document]
brozorec: need more info on what this controls

brozorec (changes requested):

DaigaroCota:
The `openVaultFactory` state variable is added  to control when the Chief.sol contract will allow or not permissionless creation of vaults.

brozorec (approved):

PR #96, opened 2022-10-20
Compound V2

brozorec:
- [x] add interfaces for cTokens
- [x] create a `CompoundV2` lending provider contract
- [x] made `BorrowingVault` payable so that it can support cETH
- [ ] tests

Note: I tried fuzzing "depositAndBorrow" but got this error
image


brozorec:
@DaigaroCota what should we put here as an operator?
https://github.com/Fujicracy/fuji-v2/blob/0ba22023e55fbef65f4200efe713cb6c245e9c93/packages/protocol/src/providers/mainnet/CompoundV2.sol#L57

DaigaroCota:
> @DaigaroCota what should we put here as an operator?
> 
> https://github.com/Fujicracy/fuji-v2/blob/0ba22023e55fbef65f4200efe713cb6c245e9c93/packages/protocol/src/providers/mainnet/CompoundV2.sol#L57

For CompoundV2 it should return the corresponding `asset` equivalent cMarket. In CompoundV2 `vault` can just be ignored.
> `function approvedOperator(address asset, address vault) external view returns (address operator);`

Recall that this is needed in the vault when `setActiveProvider()`:
https://github.com/Fujicracy/fuji-v2/blob/a2aa239317274fae15aea580a30a58ca5ce7494d/packages/protocol/src/abstracts/BaseVault.sol#L490

DaigaroCota (changes requested):
Hey @brozorec, the following test: `test_paybackAndWithdraw` for `~packages/protocol/test/mainnet/compoundV2/Provider.t.sol` is not passing.

image

This is because when you set the `ActiveProvider` by calling `setActiveProvider()` the BorrowingVault is not giving ERC20-approval for the `debtAsset`. This means that in BorrowingVault we need to override the `setActiveProvider()` function and grant infinite approval for both the asset and debtAsset. This could be a bit tricky given that the signature of ILendingProvider changed a bit for `approvedOperator()`.

DaigaroCota (approved):

PR #100, opened 2022-10-21
OZ Timelock

brozorec:
My proposal for using directly OZ "TimelockController" instead of reimplementing it ourselves. 

@DaigaroCota you named very well the contract that's listing all the roles "CoreRoles". I think timelock roles are not "core" ones, they are equally important but they belong to an external module. That's why if we remove "TIMELOCK_*" roles from "CoreRoles" nothing else prevents us to use directly OZ's implementation.


DaigaroCota (approved):
Very good sir.

PR #152, opened 2022-11-28
Morpho as Lending Provider

pedrovalido:
**Adding Morpho as a Provider**

Morpho is a **layer on top of Aave and Compound** that matched Lender/Borrowers p2p to boost their ratings. The initial idea was to to MorphoAave and MorphoCompound and then merge them as a single provider Morpho. This doesn't seem to be possible for a some reasons:
 - they have **different rates** and there's no way of rebalancing to morpho without deciding between MorphoAave and MorphoCompound
 - they have **different functions** regarding their interface
 - they have **different addresses** of interfaces to communicate with Aave or Compound
 
Regarding the **addrmapper**, it was necessary to use it because morpho r**equires us to input the cToken/aToken** when depositing, withdrawing, etc. So, I added the address mapping of the tokens used and corresponding cTokens/aTokens. An **issue** about the standard way of using the AddrMapper arose. Refer to the discord thread about it to discuss


pedrovalido:
For Morpho team:

This is the PR where both Morpho Aave and Morpho Compound have been integrated. This integration consists of three types of files: 
- interfaces we use from Morpho (ILens and IMorpho)
- contracts implementing the ILendingProvider interface from Fuji (where the magic happens)
- tests

### Interfaces

We imported Morpho's interfaces to be able to use the protocol's functions and interact with it.

### ILendingProvider

Fuji's interface that defines a provider and the main functions that are going to be used:
- providerName
- approvedOperator
- deposit
- borrow
- withdraw
- payback
- getDepositRateFor
- getBorrowRateFor
- getDepositBalance
- getBorrowBalance
(full description of this functions can be found here https://github.com/Fujicracy/fuji-v2/blob/main/packages/protocol/src/interfaces/ILendingProvider.sol )

### MorphoAaveV2 and MorphoCompound (contracts that implement ILendingProvider interface)
MorphoAaveV2 can be found here https://github.com/Fujicracy/fuji-v2/pull/152/files#diff-eff547687896ca89753c933cfb225dc68c41c0939e5ec466ecb177a62f7235e8
MorphoCompound can be found here https://github.com/Fujicracy/fuji-v2/pull/152/files#diff-45ee538786ad0643536b735c01fb57529b2ff3ce2b4fc4589ea076262dba8825

Both these contracts define how Fuji will interact with Morpho

### Tests

To test the integration of Morpho some tests have been run (forge test --match-contract Morpho):
- depositAndBorrow
- paybackAndWithdraw
- getBalances
- getInterestRates

packages/protocol/test/forking/mainnet/MorphoAaveV2.t.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: UNLICENSED
```
MerlinEgalite, line 1: I would recommend using precise assertion instead of GT or LT assertions

Rubilmax (commented):
commented a few things to complete Merlin's review

Issue #41, opened 2022-09-08
Borrow: Manage open position


iafhurtado:
**Meta**

Estimation Time: 5 days

[View design >](https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=509%3A13304&t=hltTwUy1et3zT5gR-1)

**Todo**
- [ ] If vault balance is null, redirect user to borrow page
- [ ] If vault balance is not null, the form display a modal that redirect to the manage position page [> design](https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=2957%3A46792&t=GMYcBLYH3XNVo9Et-1)
- [ ] I can see the icon (1) and on hover the text "Your position is currently on {name} network"
- [x] I cannot edit the token
- [ ] I can see a button add positon (2)
  - [ ] On click switch between "deposit and borrow" and "payback and withdraw" mode
- [ ] I can change the collateral amount
  - [ ] On change, the overview pane changes according to the design (3)
- [ ] I can change the debt amount
  - [ ] On change, the overview pane changes according to the design (4)
- [ ] The button is updated depending on debt and collateral, and if we are in "deposit and borrow" or "payback and withdraw" mode
- [ ] The user can either deposit || && borrow, repay || && withdraw (need to be better defined but in code it's the exact same call to sdk. What changes is how we display it.
- [ ] Actions pending display "1 pending" instead of the address (same as borrow)

**How to**
Retrieve balance of user in a vault ?
Use `BorrowingVault.getBalances(address)`. If balance is 0. then the user don't have position -> open position form
Else -> manage position form

**Others**
- Inconsistency in design: "Add collateral + Borrow => Deposit and borrow"
- Missing: what should we do on cross chain tx ? See @brozorec comment below for a full explanation.

ferostar:
@brozorec Guillaume closed his [PR](https://github.com/Fujicracy/fuji-v2/pull/187) with the message `Cancelled and closed. Store refacto can be found on https://github.com/Fujicracy/fuji-v2/pull/280`. I guess the work was outdated and it wasn't worth continuing it. Meaning it needs to be done from scratch on a fresh component instead of separating `Borrow` into two.

brozorec:
@ferostar yep, that's correct. As we said, he started implementing it on the same components as the borrow page but we wanted to separate them.

ferostar:
[...]

PR #181, opened 2022-12-09
Markets: populate data

### LINKED ISSUE #155: Markets: populate data
### Context
After integration, we need to populate the data of fuji markets.

### To do

- [x] Fetch can takes time, therefore display a loader when we don't have data
- [x] Borrow & Collateral
- [x] Chain with the best rate
- [x] APY & APR
- [x] Waterdrop rewards
- [x] Integrated protocol
- [x] Available Liquidity
- [x] When user clicks on a row, he navigate to borrow page with pre filled data

### Explanations

**How to group pairs ?**
Basically, use `sdk.getAllVaultsWithFinancials()` that retrieve a list of vault. Then group vault by pair, and group by chain. Sort by APY.

A row is a Vault and represents a pair. First unfold display this pair on all available chains. Second unfold display the provider available for this vault.

**Borrow & Collateral**
Can be found inside a Vault (`Vault.debt & Vault.collateral`)

**Chain with the best rate**
Simply display the chain with the best borrow APR on this pair

**Supply APY & Borrow APR**
In the response of `sdk.getAllVaultsWithFinancials()`

The waterdrop can be both on Supply APR and Borrow APR.
On hover the waterdrop, display a tooltip with 
```
XX.XX% borrow APR (base) - XX.XX% reward

XX.XX% supply APY (base) - XX.XX% reward
```

**Available Liquidity**
In sdk.getAllVaultsWithFinancials()

**Integrated protocol**
Can be retrieved with `vault.getProviders()`

**User click on row**
Use useStore.changeActiveVault(vault) to set the vault the navigate.

### Related
- ~~APR **waterdrop** will be done on #140~~ moved in this ticket because it's in the response of `sdk.getAllVaultWithFinancials()`
- Saftery ratings will be done on #141
- ~~Available liquidity will be done on #136~~ moved in this ticket because it's in the response of `sdk.getAllVaultWithFinancials()`

Chris-CHEVALIER:
Close #155


brozorec:
Good job :+1: 

When one hovers over the providers' logos, can we display a tooltip with their names?

Wdyt @Markoyw ?

brozorec:
> Ok so I think this is ready for review, just need help of either @DaigaroCota or @brozorec about navigation after clicking on a Market row.

It has to navigate to "Manage position", ref https://github.com/Fujicracy/fuji-v2/issues/41

Markoyw:
Hey @brozorec @doliG 

1. That's a good idea. I am attaching an example of the tooltip design here
[https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=3042%3A51874&t=bFPZHGdDkTO3x5i9-1]

Also, I'm not sure if we are able to stack the tooltip as such:
Screenshot 2023-01-02 at 8 00 49 PM

2. Ah, it was a comment on the sorting functionality. I was trying to test it - when I click on the header, the arrow changes up to down. However, the data is not sorted.

brozorec:
Awesome, thank you @Markoyw :+1:
Regarding "2." I meant something different: it was about the order of the columns. For the assets' names, we start with the borrowed one while for the rates the supply rate is first.

Markoyw:
@brozorec 

Good point. What do you think if we were to switch the assets column starting from - 'collateral' and then 'borrow'?

doliG:
@Markoyw Design feedbacks on figma are fixed, please check again :)

1. I don't think nested tooltip are a good idea in terms of UX (not mobile friendly) + it's difficult to handle technically, but it's doable. I suggest instead to display one big tooltip with the logo of all protocols and their name.

2. Regarding the column order, I had the same remark as @brozorec and it's an easy change. @Markoyw if you're okay with it I'll do it.

3.Regarding sorting, it's because sorting the data will cause everything to change (even chain with the best rate). @brozorec can we discuss it through a call ? I think it'll be easier.

brozorec:
> @brozorec
> 
> Good point. What do you think if we were to switch the assets column starting from - 'collateral' and then 'borrow'?

@Markoyw Yep, I think it's safe to do so :+1:

brozorec:
1. Initially, we won't have vaults with more than 4 providers, so I suggest we just put a simple tooltip with the names.
2. 
3. Regarding the sorting, I'm creating a ticket for it so that we deal with it later on.

@doliG

doliG:
1. Tooltip done
2. 
3. Ok, I commented the sorting code
4. [NEW] I created #219 that can be done once we have a design. Thanks :)

@brozorec @Markoyw this PR is done, I'm merging it. Thanks for your contributions guys 🔥

doliG:
1. Ok done
2. .
3. I commented the code that sort the table, we'll handle it later
4. [NEW] #219 (waiting for design)

@Markoyw @brozorec I'm merginf this one, thanks you guys for your contributions.

PR #210, opened 2022-12-23
Vaults lack of slippage protection

DaigaroCota:


packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -34,6 +34,10 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
   error BaseVault__redeem_invalidInput();
   error BaseVault__setter_invalidInput();
   error BaseVault__checkRebalanceFee_excessFee();
+  error BaseVault__depositSlippageProtection();
+  error BaseVault__mintSlippageProtection();
+  error BaseVault__withdrawSlippageProtection();
+  error BaseVault__redeemSlippageProtection();
```
brozorec, line 40: Rename according to the convention. I'd suggest for the error description to use "slippageTooHigh", for ex. `BaseVault__deposit_slippageTooHigh()`

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -46,6 +46,8 @@ contract BorrowingVault is BaseVault {
   error BorrowingVault__rebalance_invalidProvider();
   error BorrowingVault__rebalance_invalidFlasher();
   error BorrowingVault__checkFee_excessFee();
+  error BorrowingVault__borrowSlippageProtection();
+  error BorrowingVault__paybackSlippageProtection();
```
brozorec, line 50: same comment as for the errors in "BaseVault"

brozorec (changes requested):
This PR deals only with slippage protection on a vault level. Should we extend it to the Router too (maybe in another PR)?

DaigaroCota:
> This PR deals only with slippage protection on a vault level. Should we extend it to the Router too (maybe in another PR)?

I recommend that we deal with router-swap slippage separately because is a complete different approach. I even think the slippage in the router should be implemented in the Swapper-types contract.

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
DaigaroCota: fixed as suggested.

packages/protocol/src/abstracts/BaseVault.sol
DaigaroCota: fixed as suggested.

brozorec (approved):
:+1:

Issue #224, opened 2023-01-05
Add version in Vaults  and setter for ratings in Chief


DaigaroCota:
Discussed in Smart Contracts meeting Jan 5, 2023

We decided to add a public state variable in the BaseVault contract that returns a version.
This version will be used to identify the revision of future deployments of vaults.

In addition, we agreed on creating a setter function for the vault ratings in the Chief contract.

DaigaroCota:
@brozorec I am thinking version to be a constant string with format v0.0.0? 
```
/**
   * @dev `VERSION` of this vault. 
   * Software versioning rules are followed: v-0.0.0
   *  MAJOR version when you make incompatible ABI changes
   *  MINOR version when you add functionality in a backwards compatible manner.
   *  PATCH version when you make backwards compatible fixes.
   */
  string public constant VERSION = "v-0.0.0";
```
what do you think?

Issue #214, opened 2023-01-02
First deposit leads to Vault DOS


DaigaroCota:
Where: 
https://github.com/Fujicracy/fuji-v2/blob/3c9c1e2490cf7ba381fc0adc1ff456dba9b6834b/packages/protocol/src/abstracts/BaseVault.sol#L23

Description:
Refer to issue described [here](https://rokinot.github.io/hatsfinance).
TL;DR 
The function `_convertToShares()` can be manipulated by depositing very small amount for the initial deposit, thus causing problems for future users (due to rounding errors) for future deposits.

Recommendation;
- Protect by minting a minimum amount of shares for first user.

Issue #192, opened 2022-12-14
CS - BaseVault out-of-date provider


DaigaroCota:
Where: 
https://github.com/Fujicracy/fuji-v2/blob/ab02d2308797577973ac358af8c7aadf973bcec7/packages/protocol/src/abstracts/BaseVault.sol#L487

Description:
The BaseVault has the setActiveProvider functions, callable by rebalancer, that allows to change the current provider. The function is called during rebalancing. It is crucial to call this function within one transaction to make the rebalancing process an atomic process. Otherwise, the vault might be used by users after the provider is changed and before the funds are moved, which would result in transactions being reverted (e.g. borrowing from a new provider before it gets collateral).

Recommendation:
Make the function internal and call it during the rebalancing process.

PR #225, opened 2023-01-05
BaseVault multiple fixes

DaigaroCota:
This PR addresses:

- the issue with the first deposit small amount that leads to DOS #214 
- add version to vaults and a setter function for vault's ratings in the Chief.sol #224 
- out of date provider when blindly calling `setActiveProvider()` #192

### DIFF
--- a/packages/protocol/src/Chief.sol
+++ b/packages/protocol/src/Chief.sol
@@ -87,14 +87,14 @@ contract Chief is CoreRoles, AccessControl, IChief {
   }
 
   function deployVault(
-    address _factory,
-    bytes calldata _deployData,
+    address factory,
+    bytes calldata deployData,
     uint256 rating
   )
     external
     returns (address vault)
   {
-    if (!allowedVaultFactory[_factory]) {
+    if (!allowedVaultFactory[factory]) {
       revert Chief__deployVault_factoryNotAllowed();
     }
     if (!openVaultFactory && !hasRole(DEPLOYER_ROLE, msg.sender)) {
@@ -102,28 +102,28 @@ contract Chief is CoreRoles, AccessControl, IChief {
     }
     _checkRatingValue(rating);
 
-    vault = IVaultFactory(_factory).deployVault(_deployData);
+    vault = IVaultFactory(factory).deployVault(deployData);
     vaultSafetyRating[vault] = rating;
     _vaults.push(vault);
 
-    emit DeployVault(_factory, vault, _deployData);
+    emit DeployVault(factory, vault, deployData);
   }
 
   /**
-   * @notice Set `vaultSafetyRating` for `_vault`.
+   * @notice Sets `vaultSafetyRating` for `vault`.
    * Requirements:
    *  - Emits a `SafetyRatingChange` event.
    *  - Only timelock can change rating.
    *  - `newRating` is in range [1=100].
    *  - `vault_` is not zero address.
    */
-  function setSafetyRating(address vault_, uint256 newRating) external onlyTimelock {
-    _checkValidVault(vault_);
+  function setSafetyRating(address vault, uint256 newRating) external onlyTimelock {
+    _checkValidVault(vault);
     _checkRatingValue(newRating);
 
-    vaultSafetyRating[vault_] = newRating;
+    vaultSafetyRating[vault] = newRating;
 
-    emit SafetyRatingChange(vault_, newRating);
+    emit SafetyRatingChange(vault, newRating);
   }
 
   /**
@@ -140,16 +140,16 @@ contract Chief is CoreRoles, AccessControl, IChief {
   }
 
   /**
-   * @notice Set `_factory` as an authorized address for vault deployments.
+   * @notice Sets `factory` as an authorized address for vault deployments.
    * - Emits a `AllowVaultFactory` event.
    */
-  function allowVaultFactory(address _factory, bool allowed) external onlyTimelock {
-    _checkInputIsNotZeroAddress(_factory);
-    if (allowedVaultFactory[_factory] == allowed) {
+  function allowVaultFactory(address factory, bool allowed) external onlyTimelock {
+    _checkInputIsNotZeroAddress(factory);
+    if (allowedVaultFactory[factory] == allowed) {
       revert Chief__allowVaultFactory_noAllowChange();
     }
-    allowedVaultFactory[_factory] = allowed;
-    emit AllowVaultFactory(_factory, allowed);
+    allowedVaultFactory[factory] = allowed;
+    emit AllowVaultFactory(factory, allowed);
   }
 
   /**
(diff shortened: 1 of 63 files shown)


DaigaroCota:
Self note:
Need to understand why these are failing:
This is a problem with Foundry most likely.
---------------------------------------------
Failing tests:

Encountered 1 failing test in test/forking/mainnet/CompoundV3.t.sol:CompoundV3ForkingTest
[FAIL. Reason: Setup failed: Contract 0x5991A2dF15A8F6A256D3Ec51E99254Cd3fb576A9 does not exist and is not marked as persistent, see `vm.makePersistent()`] setUp() (gas: 0)

Encountered 1 failing test in test/forking/polygon/DForcePolygon.t.sol:DForcePolygonForkingTest
[FAIL. Reason: Setup failed: Revert] setUp() (gas: 0)

Encountered 1 failing test in test/forking/polygon/FlasherAaveV3.t.sol:FlasherAaveV3ForkingTest
[FAIL. Reason: Setup failed: Revert] setUp() (gas: 0)

Encountered 1 failing test in test/forking/polygon/WePiggyPolygon.t.sol:WePiggyPolygonForkingTest
[FAIL. Reason: Setup failed: Revert] setUp() (gas: 0)

DaigaroCota:
@brozorec this is ready for review. Please see notes above.

brozorec:
Hey @DaigaroCota I just had a look at this PR. I think I understand why you brought back the `debtAsset` in the BorrowingVault. It's because it's needed in `setProviders`, right? If this is the only reason, I might suggest we turn `_setProviders(providers)` into an abstract method in BaseVault. That will allow the BorrowingVault to implement it by using both `_asset` and `_debtAsset` w/o checking nullity and the YieldVault only using `_asset`.

brozorec:
Regarding the ratings, I suggest we switch from a string-based rating to a number-based one (1-100). That's easier to handle and far more scalable, at least from an SC perspective. 

How this number gets interpreted on the client side is another question. We can display whatever we want: "A+", "A-", ... or 10-base or 20-base. A rating with a range 1-100 gives us enough flexibility.

DaigaroCota:
> Regarding the ratings, I suggest we switch from a string-based rating to a number-based one (1-100). That's easier to handle and far more scalable, at least from an SC perspective.
> 
> How this number gets interpreted on the client side is another question. We can display whatever we want: "A+", "A-", ... or 10-base or 20-base. A rating with a range 1-100 gives us enough flexibility.

Agree.  I will simply assign it a uint256, and restrict range [1,100]

DaigaroCota:
@brozorec, this should be ready to merge once approved.

packages/protocol/src/Chief.sol
```diff
@@ -86,7 +89,7 @@ contract Chief is CoreRoles, AccessControl, IChief {
   function deployVault(
     address _factory,
     bytes calldata _deployData,
-    string calldata rating
+    uint256 rating
```
brozorec, line 92: Inconsistent naming, why do we use here "_name"? There are no global `factory` or `_deployData`

packages/protocol/src/Chief.sol
```diff
@@ -97,12 +100,32 @@ contract Chief is CoreRoles, AccessControl, IChief {
...
 
+  /**
+   * @notice Set `vaultSafetyRating` for `_vault`.
+   * Requirements:
+   *  - Emits a `SafetyRatingChange` event.
+   *  - Only timelock can change rating.
+   *  - `newRating` is in range [1=100].
+   *  - `vault_` is not zero address.
+   */
+  function setSafetyRating(address vault_, uint256 newRating) external onlyTimelock {
```
brozorec, line 120: In "deployVault" above, the naming is " underscore + name" while here it becomes "name + underscore"

packages/protocol/src/Chief.sol
```diff
@@ -217,4 +240,33 @@ contract Chief is CoreRoles, AccessControl, IChief {
...
+  function _checkRatingValue(uint256 rating) internal pure {
+    if (rating == 0 || rating > 100) {
+      revert Chief__checkRatingValue_notInRange();
+    }
+  }
+
+  /**
+   * @dev reverts if `vault_` is not in `_vaults` array.
+   */
+  function _checkValidVault(address vault_) internal view {
```
brozorec, line 256: some comment as above

brozorec (changes requested):

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -39,13 +40,24 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
-  IERC20Metadata internal immutable _asset;
+  /**
+   * @dev `VERSION` of this vault.
+   * Software versioning rules are followed: v-0.0.0
+   *  v-MAJOR.MINOR.PATCH
+   *  MAJOR version when you make incompatible ABI changes
+   *  MINOR version when you add functionality in a backwards compatible manner.
+   *  PATCH version when you make backwards compatible fixes.
+   */
+  string public constant VERSION = "v-0.0.0";
```
brozorec, line 51: Why not just "0.0.0"?

It's a bit weird all zeroed; maybe we start with "0.0.1"?

packages/protocol/src/Chief.sol
DaigaroCota: Updated variable naming with no underscore since you are right; there is no variable collision.

packages/protocol/src/Chief.sol
DaigaroCota: Updated.

packages/protocol/src/abstracts/BaseVault.sol
DaigaroCota: updated.

DaigaroCota:
@brozorec review comments have been addressed.

brozorec (approved):

packages/protocol/test/forking/arbitrum/AaveV3.t.sol
```diff
@@ -16,14 +16,14 @@ contract AaveV3ArbitrumForkingTest is Routines, ForkingSetup {
   uint256 public constant BORROW_AMOUNT = 200 * 1e6;
 
   function setUp() public {
-    deploy(ARBITRUM_DOMAIN);
+    setUpFork(ARBITRUM_DOMAIN);
 
     aaveV3 = new AaveV3Arbitrum();
     ILendingProvider[] memory providers = new ILendingProvider[](1);
     providers[0] = aaveV3;
+    vm.makePersistent(address(aaveV3));
```
pedrovalido, line 24: Why do we have vm.makePersistent in this test and not the other ones?

DaigaroCota: > Why do we have vm.makePersistent in this test and not the other ones? @DaigaroCota

@pedrovalido I forgot to remove it. Please do so in one of your provider implementations. It should not be affecting test passing.

pedrovalido: Got it will do. It doesnt affect, I was just curious 😅

PR #244, opened 2023-01-12
0vix as lending provider

pedrovalido:


pedrovalido:
0vix is a fork of compound with some differences. Their version of calculating rates is based on block timestamp (vs block numbers like on Compound) . Therefore, their function names on their ICToken are named accrualBlockTimestamp, borrowRatePerTimestamp and supplyRatePerTimestamp.
I created a ICToken interface with these functions specific to 0vix (plus the ones we need for the rest of the integration).
It was also necessary to :
 - create a different implementation of LibCompoundV2 (LibOvix) to reflect this changes
 - change the way we're calculating the borrow/supply rates to calculate using number of seconds per year instead of number of blocks per year

DaigaroCota (approved):

PR #249, opened 2023-01-12
Agave as Lending Provider

pedrovalido:
Integrating agave lending provider on gnosis chain


pedrovalido:
Gnosis chain was already configured on protocol/feat/hundred-lending-provider so I merged into this branch to use it.

DaigaroCota (approved):

PR #252, opened 2023-01-14
Protocol/fix/cs swapper slippage too high

DaigaroCota:
This PR should be merged after 225.


DaigaroCota:
@brozorec this is the last of the CS - security audit findings. I implemented 3 simple tests for the swapper. 

In the process of fixing the swapper, I thought that we should implement quoting methods, wdyt?  

I implemented these `compute_amountOut` and `compute_amountIn` in the test file, but I actually think this should be part of the Swapper interface itself. 

https://github.com/Fujicracy/fuji-v2/blob/44f3cbd919005a9cebc13978a5ab271d549efeb0/packages/protocol/test/forking/mainnet/UniswapV2Swapper.t.sol#L84-L115

In addition, I propose we make an additional ticket-issue to make the swapper more secure against slippage. This is by checking the price against an external oracle. We did something similar on V1.
 
https://github.com/Fujicracy/fuji-v1/blob/c040d6407111be1a48e54fb847856fe2ce80ff6a/packages/hardhat/contracts/mainnet/Fliquidator.sol#L461-L476

DaigaroCota:
@brozorec  ready for review. 

Additional issue created to add external oracle check; refer to #255 .

DaigaroCota:
@ComposableSecurityTeam this is 2/2 missing PRs to complete Audit findings (excluding informational).

brozorec (approved):

PR #257, opened 2023-01-17
Balancer flashloan provider

pedrovalido:
Integrating Balancer as Flashloan provider in every chain we currently have. Turns out they have it deployed on mainnet, arbitrum, optimism, polygon and even gnosis! Btw, big plus their fee being 0 cost which means after this PR, we will have a flasher on every chain with 0 cost :)
I took advantage of the tests done previously for FlasherAaveV3 (which include a super simple flashloan and a rebalance with Rebalancer Manager which is the real use case scenario)

### DIFF
--- a/packages/protocol/src/flashloans/FlasherBalancer.sol
+++ b/packages/protocol/src/flashloans/FlasherBalancer.sol
@@ -11,6 +11,7 @@ import {BaseFlasher} from "../abstracts/BaseFlasher.sol";
 import {IBalancerVault} from "../interfaces/balancer/IBalancerVault.sol";
 import {IFlashLoanRecipient} from "../interfaces/balancer/IFlashLoanRecipient.sol";
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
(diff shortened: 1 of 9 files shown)


DaigaroCota:
@pedrovalido  I will review this tomorrow. 

Only a non-PR related comment, I believe balancer at some point introduced a small fee on eth-mainnet for their flashloans? This was not the case anymore? Just double check. It has been months last time I checked.

pedrovalido:
From the Balancer's vault we can see the flasloan fee percentage is returned by the IProtocolFeesCollector's function getFlashLoanFeePercentage(). Both that function and the IProtocolFeesCollector interface have been added.
Using them in the computeFlashloanFee inside Fuji's FlasherBalancer it would be simple (and slightly incorrect) to just return fee * feePercentage (feePercentage being the value returned by the function mentioned above).
From balancer's calculation of the fee to be paid
Screenshot 2023-01-18 at 22 07 28
They are using FixedPoint.mulUp function:
Screenshot 2023-01-18 at 22 08 56

Issue #262, opened 2023-01-19
General documentation


DaigaroCota:
Preparing for code4rena:

- [x] IRouter and routers derivatives -> @brozorec
- [x] IChief -> @brozorec  
- [x] IVault and derivatives, factories, vaultpermissions -> @DaigaroCota 
- [x] ISwapper -> @DaigaroCota 
- [x] ILendingProvider and derivatives, including related provider interfaces and libraries -> @pedrovalido
- [x] IFlasher and derivatives -> @pedrovalido
- [x] AddrMapper -> @pedrovalido

brozorec:
@DaigaroCota @pedrovalido I started reviewing the comments in Chief. Plz check carefully the changes and let me know if this format works for you and what we can improve. 

Throughout the whole code base, we use different ways to comment, different **punctuation**, there are missing natspecs for some functions, etc.

After we validate the format, we MUST stick to the same one. If we fail to do so, it's only a sign of negligence and bad coding practices.

DaigaroCota:
I reviewed and it looks ok on my side. 

The question I have now is:

- Should we all commit in the same branch ? `protocol/docs/review-code-comments`
- Should we comment in a similar fashion the internal functions? In this case I recommend we focus the comments as `@dev` instead of `@notice`. From what I recall `@notice` is intended for public\external functions.

pedrovalido:
**Should we all commit in the same branch ? protocol/docs/review-code-comments?**
I say yes in the same branch as we all have different files do it shouldnt be a problem to do so

**Should we comment in a similar fashion the internal functions? In this case I recommend we focus the comments as @dev instead of @notice. From what I recall @notice is intended for public\external functions.**
Yes, let's do a standard for us: @notice for external functions and @dev for internal function. lmk what you think

brozorec:
- I suggest we work on different branches and PRs to easier review each other.

- I agree on the usage of @dev and @notice, I wasn't aware of the difference in their usage.

brozorec:
```
@param assets Amount of assets to deposit.
```
or
```
@params assets amount of assets to deposit.
```
Which one should we use? @DaigaroCota @pedrovalido

pedrovalido:
I would say the bottom one:
`@params assets amount of assets to deposit.`

pedrovalido:
[...]

PR #265, opened 2023-01-20
Improve code comments in Chief and Routers

brozorec:
tackles the first two items in #262


packages/protocol/src/Chief.sol
```diff
@@ -110,37 +148,45 @@ contract Chief is CoreRoles, AccessControl, IChief {
   }
 
   /**
-   * @notice Sets `vaultSafetyRating` for `vault`.
+   * Sets `vaultSafetyRating` for `vault`.
```
DaigaroCota, line 151: This should have a @notice ?

brozorec:
@DaigaroCota I think this is ready for review

packages/protocol/src/Chief.sol
```diff
@@ -22,12 +25,60 @@ import {CoreRoles} from "./access/CoreRoles.sol";
 contract Chief is CoreRoles, AccessControl, IChief {
   using Address for address;
 
-  event OpenVaultFactory(bool state);
+  /**
+   * @dev Emitted when the deployements of new vaults is alllowed/disallowed.
+   *
+   * @param allowed `true` to allow, `false` to disallow
```
pedrovalido, line 31: Not sure if true is variable and needs to have "`". Im fine either way btw.

packages/protocol/src/Chief.sol
```diff
@@ -22,12 +25,60 @@ import {CoreRoles} from "./access/CoreRoles.sol";
...
+   * @dev Emitted when a new vault gets deployed.
+   *
+   * @param factory address of the factory through which to deploy
+   * @param vault address of the newly deployed vault
+   * @param deployData encoded args used to deploy the vault
+   */
   event DeployVault(address indexed factory, address indexed vault, bytes deployData);
+
+  /**
+   * @dev Emitted when `_vaults` are set through `setVaults()`.
```
pedrovalido, line 45: should we change to {IChief-setVaults} ?

packages/protocol/src/Chief.sol
```diff
@@ -39,13 +90,20 @@ contract Chief is CoreRoles, AccessControl, IChief {
   error Chief__checkRatingValue_notInRange();
   error Chief__checkValidVault_notValidVault();
 
+  /**
+   * @dev When `permissionlessDeployments` is 'false', only addresses with this role
```
pedrovalido, line 94: same comment regarding the "`"

packages/protocol/src/Chief.sol
```diff
@@ -78,14 +152,56 @@ contract Chief is CoreRoles, AccessControl, IChief {
...
+    address[] memory previous = _vaults;
+    delete _vaults;
+    _vaults = vaults;
+    emit SetVaults(previous, _vaults);
   }
 
+  /**
+   * @notice Set `permissionlessDeployments`.
+   *
+   * @param allowed anyone can deploy a vault when `true`,
```
pedrovalido, line 177: same comment regarding "`"

pedrovalido (commented):
Nice work 😀 50 files ...

Not sure if we should be spending more time changing the details mentioned. I found some more throughout the pr. Btw Im fine if we merge this like this.

packages/protocol/src/Chief.sol
brozorec: I thought this notation was to be used only when referring to another contract. Here `setVaults()` is a method from the same one. But I might be wrong, @DaigaroCota can you confirm plz?

packages/protocol/src/Chief.sol
brozorec: That's right, will change to "true" and "false"

packages/protocol/src/Chief.sol
```diff
@@ -78,14 +152,56 @@ contract Chief is CoreRoles, AccessControl, IChief {
...
+   *
+   * @dev This method should only be used in extraordinary cases.
+   * Requirements:
+   *  - Must be called from a timelock.
+   */
+  function setVaults(address[] memory vaults) external onlyTimelock {
+    address[] memory previous = _vaults;
+    delete _vaults;
+    _vaults = vaults;
+    emit SetVaults(previous, _vaults);
```
brozorec, line 171: btw @DaigaroCota did you add this function recently and what's the rationale behind it?

brozorec:
> Nice work grinning 50 files ...
> 
> Not sure if we should be spending more time changing the details mentioned. I found some more throughout the pr. Btw Im fine if we merge this like this.

Ohh I modified only four files; the rest is from Daigaro's merging :smile:

packages/protocol/src/Chief.sol
DaigaroCota: I am having problems with the test for factory deployment. In particular, it fails when trying to deploy a BorrowingVault, but I think it is a problem with Foundry, since the same code was working before. Anyhow, to test vault pausability from the Chief, all vaults must be deployed using the `deployVault()` method and I realized this was a problem. There is no other way to record vaults in the {Chief} other than using this method. Considering a case in where we have to "clean" obsolete vaults (for example we uprades or so), we will require such a function.

DaigaroCota: For that reason, I refactored the vault tests so that we are isolating the functionalities we are testing. The pausability tests and the rebalancing tests were using the deploy factories and were failing because of this Foundry problem (perhaps is something else but I still need to figure it out).

packages/protocol/src/Chief.sol
DaigaroCota: `setVaults()` is a method in the same Chief contract. Therefore there shouldn't be the need to self refer to it as {Chief-setVaults}.

packages/protocol/src/Chief.sol
DaigaroCota: Update. I found my error. I commented a line in LibSSTORE2 that I should not had. Deployments from factory now work, and we have cleaner refactored vault tests. 👯

DaigaroCota (approved):

packages/protocol/src/Chief.sol
DaigaroCota: I was assumed it was not needed because of this statement:
https://github.com/Fujicracy/fuji-v2/pull/267#issuecomment-1399598921

PR #267, opened 2023-01-21
Improve code comments in Vault and Swapper 

DaigaroCota:


DaigaroCota:
@brozorec  during my review of the documentation of the Vault and derivate contracts, I realized that the abstract contract {VaultDeployer} implements a function called `_registerVault()`.

This function registers the vaults merely based on asset, however, it makes no reference to map or identify the `debtAsset`. Is this a feature or a bug? I am not sure how the SDK is "populating" or creating the list of available vaults, but just wanted to share that at this moment is will probably take a secondary rpc call to determine the `debtAssets` of the vault. 

https://github.com/Fujicracy/fuji-v2/blob/8f9ecf5b711604a1bb94d208c8a08541634fcfcc/packages/protocol/src/abstracts/VaultDeployer.sol#L47-L53

brozorec:
@DaigaroCota I need to check this; we don't use it for the moment (so the question would be "should we keep it at all"). The VaultDeployer registers the vault in a generic way and it's on the factory to store them according to their specifics.

PR #278, opened 2023-02-02
Update readme of frontend

doliG:
Hey @ferostar ,

I try to give and overview of the intention I build the frontend with, I let you check and give me some feedback.

If you need you can reach me on discord :`yome#9166`

Thanks :)


ferostar:
Thanks @doliG , started reading the updates as soon as I saw the commit 😁.

doliG:
Haha so fast ! Nice, tell me if anything is unclear or if there are typo, I need to go back on this anyway think of this PR as a draft ;) But rn it's 10:30pm here so I'll continue tmr

ferostar:
@doliG I'm surprised about zustand's growth these last couple of years. While I haven't used it, it can't be any worse than mobx 😅 -the thread you shared got a bit too philosophical for my tastes. The developer wasn't much of help, IMHO. Anyway, some questions:

- I wanted to ask you about store creation since there's some commented code and deprecated functions so I'm not sure if that's a wip or if the code had some issues.
- Using `storybook` might be a good idea! I don't like much having `Theming` in the middle of all other components, feels... dirty lol
- I do like the project structure. Agree with the modular approach! I wonder about the difference between `/` and `Layout`.  I'm sure some components will fight both.
- You say `we're still at a very early stage so I don't think we need to but too much focus on having a design system. Better make it work first and refacto later.` While I understand spending much time on this is a waste and a luxury we might not have, I don't think we can just let it be for the time being. This is a moment in which refactoring is still doable. Once we have all of the pages, refactoring is going to be a nightmare -it's going to take a lot more time then, and the development process will be (a bit) slower (I think).

doliG:
Hey @ferostar 

- **About store creation**

So there are few things here. A refacto of the store is in progress. Basically I'm moving each slice into a separate store, because zustand typing makes build fail if you combine slice + typing + devtools.

Regarding the depreciation of `create`, it's just a matter of import, it seems like they changed the way it works since I write our store. We need to move from `import create from "zustand"` to `import { create } from "zustand"`.

I'm currently fixing this in PR #280 

- **Storybook**

Well, it's up to you. I agree that a storybook would be great and very useful for mark and FE dev to work together. And yes the "theming" page is a little weird, we wrote this thinking of it like wordpress theme (don't laugh 😭 ), and it was better than nothin. But indeed using storybook can take it ten steps further. If you think it's worth it then please go :)

- **Difference between `/` and `Layout/`**

At first Layout was supposed to contains components such as `Header`, `Footer` etc... but soon we realised that we had a lot of components that doesn't really fit as "layout components". Renaming `Layout/` into `Shared/` may be a better way 🤔  

- **Regarding refactoring**

Yes I totally agree with you. In that point I failed to simply say "hold on we need to refacto", I focused too much on shipping, that's my bad. If you need help on this matter, we can have a call :)

ferostar:
@doliG thanks for the work!

A couple of points

### Project structure

- Renamed `/Layout` to `/Shared`
- Moved all Icon components to `/Shared/Icons`
- Added `index.ts` to `/Shared/Icons` to easily use them across the app
- Moved `getTokenImage` from `TokenIcon` to a new service
- There are still a couple of components in `/`, now that it’s `/Shared` I think they should probably be moved there?

### Store 

Made a few changes to the `README` file, to make it a bit more generic. Added a few things about the store (currently in https://github.com/Fujicracy/fuji-v2/pull/280). We can still add some more info and we’ll have to update the svg file.

doliG:
Hello @ferostar , I read your changes and def agree to it. Thanks for fixing typos along the way :)

> There are still a couple of components in /, now that it’s /Shared I think they should probably be moved there?
Yep I agree

brozorec:
@ferostar lmk when we can merge this PR

Issue #283, opened 2023-02-07
Lending Proivders and Assets Logos


brozorec:
Hey, these are the lists of the lending providers and assets that will use at launch. Can we make sure we have their logos available in Figma and in the dedicated FE directory? @ferostar @Markoyw :pray: 

### Lending providers
1. Aave V2
2. Aave V3
3. Compound V2
4. Compound V3
5. Euler
6. Morpho
7. Hundred
8. WePiggy (WPC)
9. dForce
10. IronBank (IB)
11. Agave
12. 0vix
13. Radiant

### Assets
1. WETH
2. WBTC
3. WMATIC
4. USDC
5. USDT
6. DAI

Markoyw:
Hey @brozorec  @ferostar 

Here's the link to the logo assets:
https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=3771%3A74579&t=6nbdpT5UuKJ9wiMB-1

ferostar:
Thanks @Markoyw 

@brozorec my account has exporting disabled, I think it's a permissions issue.

brozorec:
Sorry but I should've included here the list of chains as well:

- Ethereum
- Polygon
- Arbitrum
- Optimism
- Gnosis
- BNB Chain

ferostar:
@Markoyw above was meant for you I think, we need those assets too 😅

Markoyw:
Hey @ferostar  @brozorec 

Sure, thanks for following up on that. 

Here's the link to the network logos:
https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=3882%3A72829&t=qxH5WBkGt6eV4nEc-1

brozorec:
done with #287 and #284

PR #284, opened 2023-02-07
Update frontend assets

ferostar:
From issue https://github.com/Fujicracy/fuji-v2/issues/283


ferostar:
@brozorec a couple of quick questions:

We have a list of `network` assets. They're

```
Arbitrum
Binance
Ethereum
Fantom
Optimism
Polygon
Potcoin
```

Which ones can we remove for v1?

Also... assets for testnets are duplicated, so we have

```
Optimism Goerli
Goerli
Mumbai
```

as well. I think we should just map testnet icons to the corresponding mainnet, otherwise we'll have duplicated assets without any need.

brozorec:
Sure, let's reuse assets, especially if it's for testnets

ferostar:
@brozorec got it. And what about the networks, can we remove any?

brozorec:
> @brozorec got it. And what about the networks, can we remove any?

Yeah, I don't even know wtf is Potcoin lol

Fantom is not supported by Connext yet but let's keep it, I think they will add it at some point.

PR #285, opened 2023-02-08
Permissionless connext xReceive

DaigaroCota:
This PR addresses the fact that previous implementation of the Connext-IReceiver was using authentication to allow call through. The idea is to make call permissionless, however, still preventing 3rd actors to access funds from failed xCalls.


DaigaroCota:
This PR is ready for review. My approach to the problem was to attack the root cause directly. We should not allowed failed Connext-xCalls to leave funds at the `ConnextRouter`.  Therefore, the logic sends all funds of failed transferIds to the `ConnextHandler`. From there we can manage this failed txs. 
@brozorec please check and let me know of any potential refactor or optimization. Variable renaming too.

brozorec:
We want to add a beneficiary check on the "XTransferWithCalldata" action as discussed during the weekly.

brozorec:
@DaigaroCota regarding https://github.com/Fujicracy/fuji-v2/pull/291#issuecomment-1427195326
and as you set a revert if "swap" is a first action, it doesn't make sense anymore `_replaceAmountInSwapAction()` in `_accountForSlippage()`, right?
If that's true and you want to remove it, then plz also update the comments.

DaigaroCota:
> @DaigaroCota regarding [#291 (comment)](https://github.com/Fujicracy/fuji-v2/pull/291#issuecomment-1427195326) and as you set a revert if "swap" is a first action, it doesn't make sense anymore `_replaceAmountInSwapAction()` in `_accountForSlippage()`, right? If that's true and you want to remove it, then plz also update the comments.

Removed `_replaceAmountInSwapAction()` function  in `_accountForSlippage()`. I also updated tests to ensure everything is passing.
We should be good to go. @brozorec  👍

brozorec (approved):

PR #286, opened 2023-02-08
Routing modal

ferostar:


ferostar:
@brozorec just did a first commit to be able to publish. I wanted to mention something. Every time we make a change that requires the frontend to update the vaults, like changing changes, etc, we do:

```
const availableVaults = await sdk.getBorrowingVaultsFor(
  collateral,
  debt
)
const [vault] = availableVaults
if (!vault) {
  // ...
  return
}
set({ availableVaults })
await get().changeActiveVault(vault)
```

As you see in the last line, we set the active vault with a value we got doing `const [vault] = availableVaults.`. The whole idea is to offer a recommended one, so.. we can just set the first element as selected, or we can have a way of choosing one. A sort of `recommended` property, something, to easily know a single vault is the recommended one instead of just the array position. Thoughts?

brozorec:
You can safely set the first vault as the recommended one. The SDK takes care of sorting them according some criteria https://github.com/Fujicracy/fuji-v2/blob/1e2ca9a4948fb1de27dcdbeb0d451c750759d65a/packages/sdk/src/Sdk.ts#L201-L209

ferostar:
@brozorec two notes

- Steps are missing for bridges, since they need to be dealt with differently. Would be nice to know the full list of steps, actually, in case there are more.

- Still need to implement changing a route, so far in polygon -> polygon I always get one vault and then one route.

brozorec:
@ferostar 
> Steps are missing for bridges, since they need to be dealt with differently. Would be nice to know the full list of steps, actually, in case there are more.

You can find a list of the steps here:
https://github.com/Fujicracy/fuji-v2/blob/main/packages/sdk/src/enums/RoutingStep.ts

Some notes for the "bridge" step:
https://github.com/Fujicracy/fuji-v2/blob/47503cec1910307de99b74b3d531879db13b8fdf/packages/sdk/src/types/RoutingStepDetails.ts#L7-L14
---

A quick feedback for what's already implemented:
image

- no need to display separately "start" and "end"
- for the bridge step, we display the logo of the destination chain
- for deposit, borrow, repay, withdraw steps, we use the token icon instead of lending provider

ferostar:
@brozorec most of the work is done, we need to test routes with bridges and multiple vaults in order to make sure everything works as expected, there might be a few unnoticed bugs

brozorec:
@ferostar submitted a review PR #310 

You can use WETH-USDC for Optimism <> Polygon to see two routes.

PR #291, opened 2023-02-11
ConnextHandler Suggestions

brozorec:
@DaigaroCota These are my suggestions about #285 

1. you talk about transfers; I think it's more accurate to call them transactions
2. removed the error from the constructor. I think if we pass the zero address, it will fail anyway because of the casting to "ConnextRouter" (I'm not super sure about that, though)
3. Why do we bother to replace `sender` in `recordFailed()`, when in `executeFailedWithUpdatedArgs()` we can pass whatever `args` we want?

### DIFF
--- a/packages/protocol/src/routers/ConnextHandler.sol
+++ b/packages/protocol/src/routers/ConnextHandler.sol
@@ -49,7 +49,7 @@ contract ConnextHandler {
   ConnextRouter public immutable connextRouter;
 
   /// @dev Maps a failed transferId -> calldata
-  mapping(bytes32 => FailedTxn) private _failedTransfers;
+  mapping(bytes32 => FailedTxn) private _failedTxns;
 
   modifier onlyConnextRouter() {
     if (msg.sender != address(connextRouter)) {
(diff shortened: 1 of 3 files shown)


DaigaroCota:
> @DaigaroCota These are my suggestions about #285
> 
> 1. you talk about transfers; I think it's more accurate to call them transactions
> 2. removed the error from the constructor. I think if we pass the zero address, it will fail anyway because of the casting to "ConnextRouter" (I'm not super sure about that, though)
> 3. Why do we bother to replace `sender` in `recordFailed()`, when in `executeFailedWithUpdatedArgs()` we can pass whatever `args` we want?

1.- Agree, lets call them transactions. I noticed that the mapping was still called `_failedTransfers`, therefore I renamed it to `_failedTxns`.
2.- ok for me. 
3.- The `sender` will always have to be replaced in order to retry executions. Therefore I was initially trying to save us having to think about making this replacement. Later, I  implemented the `executeFailedWithUpdatedArgs` in where I considered that failed txns may require other arguments to be fixed too. So in summary, I think it is ok to remove the sender replacement and just do it all in the `executeFailedWithUpdatedArgs`. I will remove it.

DaigaroCota (approved):

DaigaroCota:
I realized having swap as a first action is an open window to attacks that we cannot handle by just checking the beneficiary. This is because the attacker can also use the sweeper who can receive the amount that wasn't swapped.

PR #298, opened 2023-02-14
Morpho lending provider (fix)

pedrovalido:
As suggested by the morpho team on the PR:

- [x] change blocks per year
- [x] use totalBalance instead of suppliedOnPool + suppliedP2P
- [x] change _getAddrmapper to _getAddrMapper

Regarding the other suggestions:

- Different order of returned outputs on ILensInterface for both MorphoAave and MorphoCompound has no impact on the code as they're not being used (changing it would require "unmerging" both interfaces and is not needed as of now)
- Precise assertion instead of GT or LT is going to be changed in the future but we have a slight problem with inaccuracies as of now


packages/protocol/src/providers/mainnet/MorphoCompound.sol
```diff
@@ -94,7 +94,7 @@ contract MorphoCompound is ILendingProvider {
     (uint256 ratePerBlock,,) = ILens(LENS).getAverageSupplyRatePerBlock(_getCToken(vault.asset()));
     //no. of blocks per year
     //convert from 1e18 to 1e27
-    rate = ratePerBlock * 2102400 * 10 ** 9;
+    rate = ratePerBlock * 2612670 * 10 ** 9;
```
DaigaroCota, line 97: @pedrovalido do we need to update this parameter in all CompoundV2-like providers? If so, would you like to just do it in this PR or do it on a separate branch-PR ?

pedrovalido: Yes, we do. I've fixed it and committed the changes :)

DaigaroCota (approved):

PR #319, opened 2023-02-16
Upgrade frontend to Next.js 13.x

ferostar:
I always noticed that we have a FOUT issue (that the default font appears for a bit before switching to our font), so the whole screen changes size. Known issue.

Didn't want to spend much time on it, but I saw that in `next 13.x` they added a new submodule to preload fonts that work particularly well with Google Fonts, as the one we're using.

I quickly upgraded to `13`, solved a few problems, and changed the way we load fonts. There's no FOUT anymore, and it seems everything looks good. Again, I spent only a couple of minutes trying this, so it could be just a starting point, but the impact is significant!

@brozorec care to have a look?


brozorec:
hey @ferostar ok for me :+1: 
Would you also update "package.json" with the new version of next and also add @next/font (CI fails because it doesn't find this module)?

ferostar:
@brozorec have one more look if you want to:
- fonts should load with no FOUT
- app should start in /markets
- tapping Fuji's logo should take to /markets

ferostar:
@brozorec just realized something went missing in the UI after upgrading, I'll let you know once I'm done

ferostar:
@brozorec  turns out there's at least one [issue](https://github.com/blocknative/web3-onboard/issues/1352) between Next 13 and web3-onboard, going to see if I can quickly solve it. Otherwise, I'll leave until after the MVP.

ferostar:
Tried both solutions provided so far:

- Upgraded node to 18.x
- Upgraded `web3-onboard`
- Wrapped the app in their provider

It doesn't work. Only setting `swcMinify` to false does. But by doing so we lose its performance improvements.

Issue #331, opened 2023-02-20
Onboarding Modal: Tutorial + Terms & Agreement


brozorec:
### Context 

When the user visits the app for the first time, they need to be presented with some "how it works" content. They also need to accept the Terms & Agreement and be warned about the risks.

### Overview

- This task consists of creating two distinct elements: "Safety Notice" (SN) and "Walkthrough" (WT).
- To know if the user has accepted the terms and was walked through the onboarding, we need to save a JSON item in the LocalStorage.
- The LocalStorage item contains the date and time on which the terms were accepted and if they went through the onboarding.
- Designs are provided in #221 

### TODOs

- [x] Create a new shared component for both elements and add it to the app root.
- [x] Add new actions and properties to "auth.store" to read from and write to LocalStorage.
- [x] On load app, read from LocalStorage and check if json element is present. If it's present, skip.
- [x] The modal appears when the user visits the app for the first time and can't be closed if the user clicks outside.
- [x] On SN, the button is disabled if the user doesn't check the box.
- [x] When the user accepts the terms in the SN, save data to LocalStorage.
- [x] After the user accepts the terms on the SN, display the WT
- [x] If the user clicks on "Skip, I will explore on .." or reaches the last step of the onboarding, store data to LocalStorage and close the modal.

PR #355, opened 2023-02-28
Safety notice and onboarding added

NikolaiYurchenko:
ticket: https://github.com/Fujicracy/fuji-v2/issues/331

**PR Notes**

- created 2 modals: one for view and accept terms of use and second is onboarding
- first can not be skipped
- second can be skipped
- local storage item is set when we interract with modals
- local storage item contains isTermsAccepted flag and date of acception and (optionally) isExploreInfoSkipped flag
- after acceptance modals shouldn't be displayed ever since


ferostar:
@NikolaiYurchenko great job, thanks for the work.

Can you make a couple of really small changes?

- Move the new components out from `/components/Shared` to `/components/Onboarding`, don’t want to make `Shared` too big and the components won't be actually used by others.
- Rename `isTermsAccepted` to `hasAcceptedTerms`
- Same with `isPreviouslyAccepted`

ferostar (approved):

brozorec:
I suggest the following improvements before merging to "main":

- make the mouse cursor a pointer over "Skip, I'll explore ..."
- make the points clickable with the mouse cursor as a pointer and navigate between steps
image
- the button "Next" becomes "Finish" at the last step
- the height of the onboarding modal stays fixed throughout the three steps

NikolaiYurchenko:
> I suggest the following improvements before merging to "main":
> 
> * make the mouse cursor a pointer over "Skip, I'll explore ..."
> * make the points clickable with the mouse cursor as a pointer and navigate between steps
>   image
> * the button "Next" becomes "Finish" at the last step
> * the height of the onboarding modal stays fixed throughout the three steps

Sure, will do

PR #368, opened 2023-03-13
Fix spend allowance overloading

pedrovalido:
This pull request addresses C-1 of Macro audit report

### DIFF
--- a/packages/protocol/src/abstracts/BaseVault.sol
+++ b/packages/protocol/src/abstracts/BaseVault.sol
@@ -187,7 +187,7 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
    * @dev Called during {ERC20-transferFrom} to decrease allowance.
    * Requirements:
    * - Must be overriden to call {VaultPermissions-_spendWithdrawAllowance}.
-   * - Msut convert `shares` to `assets` before calling internal functions.
+   * - Must convert `shares` to `assets` before calling internal functions.
    *
    * @param owner of `shares`
    * @param operator allowed to act on `owners` behalf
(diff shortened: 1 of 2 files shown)


packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -205,6 +205,20 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+   * Requirements:
+   * - Must be overriden to call {VaultPermissions-_spendWithdrawAllowance}.
+   * - Must convert `shares` to `assets` before calling internal functions.
+   *
+   * @param owner of `shares`
+   * @param receiver to whom `shares` will be spent
+   * @param shares amount to spend
+   */
+  function _spendAllowance(address owner, address receiver, uint256 shares) internal override {
+    _spendWithdrawAllowance(owner, receiver, receiver, convertToAssets(shares));
```
abhishekvispute, line 219: Fuji contracts allow users to increase/decrease their withdrawal allowance per operator, per receiver through various methods of vault permissions.
https://github.com/Fujicracy/fuji-v2/blob/efed33417d3d0245673a57133ec73e7b05363696/packages/protocol/src/vaults/VaultPermissions.sol#L94-L108

Hence it's not always given that ``receiver == operator`` 
Which contradicts this PR 

If an allowance is set for a different receiver than the operator, this pr would always spend the allowance for the same receiver only.
Either consider restricting operator= receiver only in all allowance methods or refractor this fix to accommodate operator!=receiver case.
One way of doing latter is overriding `transferFrom` method directly and using `_spendWithdrawAllowance` in place of `_spendAllowance`
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/1642b6639b93e3b97be163d49827e1f56b81ca11/contracts/token/ERC20/ERC20.sol#L160

owner is from
spender is operator
to is receiver
amount needs to be converted to assets

abhishekvispute:
Next PR Made to resolve issues of this PR: https://github.com/Fujicracy/fuji-v2/pull/536

Issue #370, opened 2023-03-13
Tooltip for signature


brozorec:
### Connext
For doing actions like borrowing or withdrawing, the user needs first to authorize the Fuji router contract to act on their behalf. This is done through EIP712-type of signature. For more details about how it works, check [here](https://www.notion.so/Himalaya-Documentation-v0-1-0-7c8edb56e6764d00a66a184e48692bf9?pvs=4#8dba481d20604959aa69b2ab536e5788) or [here](https://medium.com/fuji-finance/what-powers-fujis-cross-chain-money-market-aggregator-743775d97d54).

This requirement can seem confusing for the end user, and they might not know why they should sign. That's why we want to display a section explaining this.

Here are the designs for it [figma >](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=2752%3A68052&t=E6qQuSMNMXB3ykC6-1).

### TODOs
- [x] On the borrow and the manage-positions page, when the user is ready with the inputs, and we have fetched from the SDK `routes` alongside the `actions`, check the `needPermit` property in the borrow store. If it's set to "true", display the section "Why do I need to sign?".
- [x] On hover "?", display a tooltip with the following text (ignore the one in the design):

"Before submitting the transaction, you need to sign a message to authorize the Fuji router contract to act on your behalf. You can learn more about it in [our docs](https://www.notion.so/Himalaya-Documentation-v0-1-0-7c8edb56e6764d00a66a184e48692bf9?pvs=4#8dba481d20604959aa69b2ab536e5788)"

### Notes
- The link in the tooltip is a temp one, we'll replace it with a gitbook, once the official  public docs are ready.
- Start a branch from "frontend/feat/manage-position" if it's still not merged to "main"

PR #373, opened 2023-03-13
Signature tooltip

NikolaiYurchenko:
https://github.com/Fujicracy/fuji-v2/issues/370


ferostar:
@NikolaiYurchenko I tried to borrow and the tooltip didn't show up and realized you made a mistake. 

You shouldn't use the `needsAllowance` function, that's not what we're trying to know. Use `borrow.store`'s `needsSignature` property instead. They're two completely separate things.

- To know if a particular operation needs the user's signature/permit we call`Sdk.needSignature(route.actions)`)
- To know if the we need to show the allowance modal we use `sdk.getAllowanceFor(token, Address.from(address))`

Actually you'll have to simplify the code. Remove `needsAllowance` function and replace with the `needsSignaure` prop.

ferostar:
> > 
> 
> that was my first sollution but seems like we additionally call sdk on every input change because we arent doing this now i believe. I'll work on this more

Yes, we are because in order to know the operation type we need to check the user's input in real-time. Let me give you an example:

- The user enters the screen to manage a position. By default, it's `DEPOSIT_AND_BORROW`.
- He enters a 1 in debt, which means we change to `BORROW`
- He enters 0.1 in collateral, which means we're back to `DEPOSIT_AND_BORROW`.
- The user changes a chain.
- Now he deletes the debt, so it's just a `DEPOSIT`
- Now he changes to `Remove position`

All of the above change the operation type and so we make a lot of calls to know if the user needs to sign the thing or not. Now, if we're still making too many calls and we can improve, great.

ferostar (approved):

PR #376, opened 2023-03-15
Fix denial of service borrowing vault

pedrovalido:


pedrovalido:
The issue:
When someone repays the vault's full debt, there's an issue when doing any other vault operation because of the convertDebtToShares function which divides by totalDebt (which is 0).

Chosen option to address this issue
Create a function correctDebt(amount) which borrows the given amount from the current provider. This will lead to an increase in the totalDebt (with precautions) taking advantage of the vault "repayer" and mitigating the issue when the totalDebt is 0.

ps.: There's still the lag between someone repaying the full vaults debt and the function being called. In this timeframe, the vault will not function properly until the correctDebt function is called.

packages/protocol/test/forking/mainnet/DenialOfService.t.sol
```diff
@@ -0,0 +1,158 @@
...
+    vm.startPrank(TROUBLEMAKER);
+    IERC20(debtAsset).safeApprove(address(aaveV2pool), BORROW_AMOUNT);
+    aaveV2pool.repay(vault.debtAsset(), BORROW_AMOUNT, 2, address(vault));
+    vm.stopPrank();
+
+    assertEq(vault.balanceOf(ALICE), DEPOSIT_AMOUNT);
+    assertEq(vault.balanceOfDebt(ALICE), 0);
+
+    uint256 maxAmount = vault.maxWithdraw(ALICE);
+    do_withdraw(maxAmount, vault, ALICE);
```
DaigaroCota, line 57: Going through the test I realized that in fact if some "troublemaker" paybacks the debt of the vault, users are now free to withdraw their assets without burning their debtShares. I think we need to discuss what behavior we expect.

pedrovalido: Yes, the time in between us calling the troublemaker paying the debt and we calling correctDebt we'll be a problem.

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -706,4 +708,28 @@ contract BorrowingVault is BaseVault {
...
   }
+
+  /**
+   * @notice In case someone repays this vault's debt directly to the lending provider,
+   * a borrow is done with the amount that was repaid so no issues regarding
+   * the relationship between debt and debt shares occurs.
+   *
+   * @param  amount to be borrowed
+   */
+  function correctDebt(uint256 amount) external onlyTimelock {
```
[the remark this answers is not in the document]
pedrovalido: You're correct. I'll fix this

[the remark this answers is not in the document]
pedrovalido: This is now fixed. Treasury is passed as an argument in the correctDebt function so funds are sent to it

packages/protocol/test/forking/mainnet/DenialOfService.t.sol
pedrovalido: As discussed, I paused the withdraw in the withdraw function and it is not very efficient but couldnt find a better way. Issue was added to the optimizations card in the board. 
Also, left a //TODO in the correctDebt function to decide if we should unpause there or not. I remember we touched this issue but didn't reach a conclusion. My suggestion: we can unpause inside the correctDebt function as it is only called by timelock so I think it is safe. Let me know if I'm missing something

DaigaroCota: Conclusion was to leave un-pausing as a separate action.

DaigaroCota (changes requested):
Hey @pedrovalido, as discussed lets proceed to remove the `amount` argument from the 'correctDebt()` function such that by default we re-establish 1 debtShare == 1 unit of debt asset. 
We will also exclude that un-pause from the correct debt function, so that this is a seperate action and the "activation of the vault in more controlled steps. In addition the OZ timelock controller contract allows to execute batch actions too. So in case of desiring an "atomic" un-pausing of the withdraw action once debt is corrected it can be done through the timelock.

pedrovalido:
> Hey @pedrovalido, as discussed lets proceed to remove the `amount` argument from the 'correctDebt()` function such that by default we re-establish 1 debtShare == 1 unit of debt asset.
> We will also exclude that un-pause from the correct debt function, so that this is a seperate action and the "activation of the vault in more controlled steps. In addition the OZ timelock controller contract allows to execute batch actions too. So in case of desiring an "atomic" un-pausing of the withdraw action once debt is corrected it can be done through the timelock.

Changes have been addressed :)

DaigaroCota (approved):

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -301,6 +303,37 @@ contract BorrowingVault is BaseVault {
...
+    internal
+    override
+    whenNotPaused(VaultActions.Withdraw)
+  {
+    uint256 totalDebt_ = totalDebt();
+    uint256 supply = debtSharesSupply;
+
+    if (totalDebt_ == 0 && supply > 0 && supply > totalDebt_) {
+      _pause(VaultActions.Withdraw);
+      revert BorrowingVault__withdraw_debtNeedsCorrection();
```
abhishekvispute, line 332: can you please explain what you intend to do here?
I see a pause on top and a revert added just below

DaigaroCota: This is the situation in where some external actor, pays entirely the debt of the vault. By that previously, they will effectively, disable the vault from borrowing operations, if I recall. 
Now, if... someone, does pay the vault's debt ever, we created a function to restore debt back, and essentially later decide what to do with the "benefactor" proceeds. 
The conditions there, is to avoid withdraws at the moment right after the debt has been paid back.

If you read the conditions: if total debt is zero, meaning... someone paidback the whole debt, and there is debtSharesSupply, meaning.... there were user/s owing funds. Then pause all withdrawals until all gets sort out.

In the case of someone paying off all the debt, we don't want to create "bad debt" position, by allowing users to withdraw the shares and leaving "unbacked" debtShares, once debt is restored.

DaigaroCota: Though the revert will do effectively the job. I see now, that state for pause will not persist.

abhishekvispute: yep general intention makes sense 

> Though the revert will do effectively the job. I see now, that state for pause will not persist.

I was more concerned with the above, yep it won't persist

DaigaroCota: Refer to #536

Issue #371, opened 2023-03-13
Placeholder for the Lend page


brozorec:
### Context
At the private launch, the lending vaults won't be available but we want to keep a reference to this feature. We have already implemented on the markets page a "Coming soon" tab for "Lending". However, clicking on the "Lend" section in the header leads to a 404 error. We want to add a similar placeholder for "/lend" as for the markets page.

### TODOs
- [x] simplify the placeholder component and comment out the connection and the email (keep only the image, the title and the message)
- [x] implement in `my-positions`, the exact same way
- [x] take the placeholder component from the markets page and bring it to the page "/lend"

![Image](https://user-images.githubusercontent.com/9572072/224763339-802308ab-460e-4136-8338-6bbe7e943ac0.png)

The text is not final and might be changed in the future.

ferostar:
Hey @NikolaiYurchenko in case you didn't notice, just added another item to the task.

PR #378, opened 2023-03-15
Lending placeholder added

NikolaiYurchenko:
https://github.com/Fujicracy/fuji-v2/issues/371

Added placeholder (image, title and text) in lending tabs on market and positions pages, and inside /lend page


ferostar:
Looks great @NikolaiYurchenko, a couple of small changes please before merging:

- Use the ternary operator `{currentTab === 0 ? <MarketsTable /> : <Lending />}`
- Move back the `Lending` screen to `/Shared`. It's a placeholder, used by different components. So we could also name it `.../Shared/LendingPlaceholder.tsx`
- Make it a bit bigger when the user enter the Lending page directly. In `Markets` and `My positions` we have more components on the screen so it looks good, but when that's the only thing we have, filling the screen a bit more would be great! Try something like `12rem` instead of `8rem`?

ferostar (approved):

ferostar:
Approved since it looks good to me, anything you want to change before merging @brozorec ?

brozorec:
> Approved since it looks good to me, anything you want to change before merging @brozorec ?

Everything looks good :+1: let's merge

PR #382, opened 2023-03-15
Big LTV warning modal

NikolaiYurchenko:
https://github.com/orgs/Fujicracy/projects/2/views/1?pane=issue&itemId=22671819

Screenshot 2023-03-15 at 20 52 46

UPD: added 2 digits after period fixation of LTV

### DIFF
--- a/packages/frontend/components/Borrow/Overview/Details.tsx
+++ b/packages/frontend/components/Borrow/Overview/Details.tsx
@@ -10,6 +10,7 @@ import { BorrowingVault, LendingProviderDetails } from "@x-fuji/sdk"
 import { NetworkIcon } from "../../Shared/Icons"
 import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined"
 import { formatUnits } from "ethers/lib/utils"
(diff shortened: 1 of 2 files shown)


ferostar:
@NikolaiYurchenko haven't tested, but just from a UI perspective, please add a couple of small changes:

- Change the amount to so it has no decimals, just `74%`. 
- Increase the font size
- Change it so there are two different paragraphs. Start the second one with "Please..."

ferostar:
@NikolaiYurchenko I'm afraid we have another issue. If the operation needs to be signed, we proceed with the action instead of showing the ltv modal:

Captura de pantalla 2023-03-16 a la(s) 20 32 41

Please make sure

NikolaiYurchenko:
@ferostar in you previous comment your LTV is 58% and modal shouldn't be shown. Can you please try in one more time because I can not reproduce

ferostar:
> @ferostar in you previous comment your LTV is 58% and modal shouldn't be shown. Can you please try in one more time because I can not reproduce

Completely right, my bad -I was looking at the recommended ltv instead of the max.

Issue #336, opened 2023-02-22
Borrow: Slippage settings


brozorec:
### Context

This ticket is the last part of #242 

When the user does a cross-chain operation, there is a chance the user suffers some slippage. More info about it, can be found [here](https://github.com/Fujicracy/fuji-v2/issues/253) and [here](https://github.com/Fujicracy/fuji-v2/issues/196#issuecomment-1425646865). We have to provide the user with a way to control the slippage settings when they are about to make a cross-chain call.

Designs for this ticket are [here](https://www.figma.com/file/v53wiSwcL5DYynVL0z6L4s/Fuji---Working-File-(NEW)?node-id=3299%3A62382&t=l5bpYVqEF49M0CSG-1).

Slippage is expressed in BPS (basis points): one basis point is equivalent to 0.01%. If the user inputs 0.50%, you should send to the dedicated sdk methods the value 50. The default slippage is 0.30% or 30 BPS.

### TODOs
- [ ] Create a new shared component and integrate it both on the borrow page and the manage position page
- [ ] The modal pops up when the user clicks on the clog icon and closes when they click on close btn or outside
- [ ] There are 3 choices between which the user can select (1%, 0.5% and 0.3%). Each of these choices is a btn.
- [ ] Selected by default is 0.3%
- [ ] The 4th element is an input, when the user clicks on the btns the input gets pre-filled. 
- [ ] The user can set their own % in the input (max 100).
- [ ] Create a new prop in the borrow store that accounts for the slippage and update it accordingly after user's interaction.
- [ ] On calling one of the SDK previews methods in `fetchRoutes()` in "packages/frontend/helpers/borrow.ts" pass the slippage as the last input (check [here](https://github.com/Fujicracy/fuji-v2/blob/main/packages/sdk/src/Previews.ts))

PR #385, opened 2023-03-16
Slippage settings

NikolaiYurchenko:
https://github.com/Fujicracy/fuji-v2/issues/336
Screenshot 2023-03-17 at 16 22 15
Screenshot 2023-03-17 at 16 22 39
Screenshot 2023-03-17 at 16 23 16


ferostar:
@NikolaiYurchenko please have a look at this issues:

- `Custom %` isn’t centered vertically, nor horizontally
- When the user starts typing a custom slippage, add a bit more left padding
- No matter what the user has selected or even typed, every time you press custom it gets back to 0.3
- You have `defaultSlippage = 30` in `SlippageSettings.tsx` but in line 26 you repeat 30 and in `borrow.store`’s initial state you hardcode 30 again. Please add `DEFAULT_SLIPPAGE` to `constants` and use every time.
- Don’t use local state for `slippage`, use the global value directly.

ferostar:
Looks great now @NikolaiYurchenko, thanks! Not merging yet until updating the `sdk`, already discussing with Boyan.

NikolaiYurchenko:
> selected the previously selected button, not the custom one.

So, I see, you want to save the previously selected value and prefill input with it. Makes sense.
Sure, I'll make this change

ferostar:
> > selected the previously selected button, not the custom one.
> 
> So, I see, you want to save the previously selected value and prefill input with it. Makes sense. Sure, I'll make this change

You should definitely do that now that you mention it, but that's not the bug I described:
- User taps slippage settings, `0.3` is selected by default. He types a custom `1.5`
- User continues doing stuff, opens the slippage settings again.
- `0.3` button appears to be selected, with a different border than the rest.

Haven't looked at the code to see if the issue is that the custom value wasn't saved or just that the UI fails to reflect it.

NikolaiYurchenko:
> @brozorec what do you think about the other point, should the screen reset to the default each time the user enters the screen or save the last value?

I think resetting to 0.3 is a better UX, but I am waiting for decision about it

brozorec:
> @brozorec what do you think about the other point, should the screen reset to the default each time the user enters the screen or save the last value?

I checked how Uniswap and Sushi deal with it, and they persist the inputted slippage over sessions. I suggest we do the same.

> you want to save the previously selected value and prefill input with it.

This is not that great. Let's not prefill the custom input with one of the selected values. Also, when the user types a custom value, it would be good to highlight it with a border.

Here is more feedback:

1. When I land on the page the slippage settings doesn't appear, I need to do an action like changing the chain in order for it to get visible.
2. When I go to the borrow or manage page, the button is in a loading state, although there's nothing to load yet. 
image
3. Design and implementation differ, icon and border-radius, also the icon is not well centered. 
image
image
4. When the dialog is open the settings button should be highlighted as in the designs.

ferostar:
> 2. When I go to the borrow or manage page, the button is in a loading state, although there's nothing to load yet.

I noticed this as well, double-checked just in case and it's not happening on the `manage-position` branch.

NikolaiYurchenko:
> Save inputted slippage across sessions (do not make it the selected value, but if the user typed 0.9, the next time he enters the webpage, the inputted value should be 0.9 instead of 0.3)

I typed 0.33, closed menu, redirected to borrow, opened it again and it is 0.33 in input and it is focused.
So if it is smth wrong with it please provide reproducing steps so i could fix unpredicted behaviour.
thank you in advance

brozorec:
> @NikolaiYurchenko if I understood @brozorec correctly, there is still one point missing:
> 
> * Save inputted slippage across sessions (do not make it the selected value, but if the user typed 0.9, the next time he enters the webpage, the inputted value should be 0.9 instead of 0.3)

Yes, that's correct. If we want to persist this value across sessions, we need to save it in the localstorage.

NikolaiYurchenko:
> Yes, that's correct. If we want to persist this value across sessions, we need to save it in the localstorage.
I see, ok

ferostar:
> > @NikolaiYurchenko now the inputted value is selected by default after reloading the screen. The custom value should be stored across sessions, but shouldn't be selected as default. @brozorec can you confirm?
> 
> Well in this case no reason to save it when we always select 0.3 as default?

Read the comment Boyan made yesterday:

> > @brozorec what do you think about the other point, should the screen reset to the default each time the user enters the screen or save the last value?
> 
> I checked how Uniswap and Sushi deal with it, and they persist the inputted slippage over sessions. I suggest we do the same.
> 
> > you want to save the previously selected value and prefill input with it.
> 
> This is not that great. Let's not prefill the custom input with one of the selected values. Also, when the user types a custom value, it would be good to highlight it with a border.
> 

The idea is:
- The user types `0.9`. We store that as custom value so the next time he enters the screen it won't show `0.3` but `0.9`
- Default value is always `0.3`

PR #389, opened 2023-03-18
Reentrancy clears checks in router

DaigaroCota:
This PR addresses C-5 in Macro report.


packages/protocol/src/abstracts/BaseRouter.sol
```diff
@@ -62,30 +60,22 @@ abstract contract BaseRouter is SystemAccessControl, IRouter {
...
-  address private _beneficiary;
+  Snapshot internal _tempTokenToCheck;
 
   /**
    * @notice Constructor of a new {BaseRouter}.
    *
    * @param weth wrapped native token of this chain
    * @param chief contract
    */
-  constructor(IWETH9 weth, IChief chief) SystemAccessControl(address(chief)) {
```
brozorec, line 83: why does the constructor need to be "payable"?

packages/protocol/src/abstracts/BaseRouter.sol
```diff
@@ -431,13 +473,14 @@ abstract contract BaseRouter is SystemAccessControl, IRouter {
...
+  function _checkBeneficiary(address beneficiary, address user) internal pure returns (address) {
+    if (beneficiary == address(0)) {
+      return user;
     } else {
-      if (_beneficiary != user) {
+      if (beneficiary != user) {
         revert BaseRouter__bundleInternal_notBeneficiary();
       }
+      return user;
     }
```
brozorec, line 484: We can avoid the nested condition so it can be re-written with "if - else if - else".

packages/protocol/src/abstracts/BaseRouter.sol
DaigaroCota: Payable functions save gas [reference](https://dev.to/zaryab2000/does-adding-a-payable-keyword-in-solidity-actually-save-gas-24bm). I implemented some subtle suggestions from the Hats findings.

brozorec (approved):

packages/protocol/src/abstracts/BaseRouter.sol
```diff
@@ -255,23 +265,23 @@ abstract contract BaseRouter is SystemAccessControl, IRouter {
         if (requestor != address(this)) {
           revert BaseRouter__bundleInternal_flashloanInvalidRequestor();
         }
-        _addTokenToList(asset);
+        tokensToCheck = _addTokenToList(asset, tokensToCheck);
 
         // Call Flasher.
         flasher.initiateFlashloan(asset, flashAmount, requestor, requestorCalldata);
```
abhishekvispute, line 271: since now the beneficiary is in memory, consider verifying requestorCalldata to check if the beneficiary passed is equal to the current beneficiary.
If not done the current fix allows attacker to have different beneficiaries between parent and child.
This was mentioned in the fix suggestion as well 
 https://www.notion.so/0xmacro/Fuji-1-Preliminary-Audit-Report-external-private-b36f908d4d3c492d8ad0a9ff9fe38b1a?pvs=4#2f4ca61ff23946b594478d72cd64144b

packages/protocol/src/abstracts/BaseRouter.sol
```diff
@@ -255,23 +265,23 @@ abstract contract BaseRouter is SystemAccessControl, IRouter {
         if (requestor != address(this)) {
           revert BaseRouter__bundleInternal_flashloanInvalidRequestor();
         }
-        _addTokenToList(asset);
+        tokensToCheck = _addTokenToList(asset, tokensToCheck);
```
abhishekvispute, line 268: hey, can you point me to a test doing actions in flash loan child ?
I have some concerns with respect to how tokensToCheck would work for assets received from flashloan

DaigaroCota: Sure, here: https://github.com/Fujicracy/fuji-v2/blob/efed33417d3d0245673a57133ec73e7b05363696/packages/protocol/test/forking/polygon/SimpleRouter.t.sol#L84

abhishekvispute: cool thanks

packages/protocol/src/abstracts/BaseRouter.sol
DaigaroCota: I agree that once the flashloan enters again it is a clean slate _beneficiary. However, the flasher calling the router back cannot contain "malicious" logic because it was checked/verified previously. 
I will drop a brainstom to check what else could be done...
Lets consider the "value extracting" actions: which those would be, withdrawing, borrowing withdrawETH, swap. Swapping cannot be the first option, because it will revert. To withdraw or borrow, you require either a permit, which is specific for an operator-receiver. Note too, that in addition the permit also now hashes the actions with the args, so permits cannot be front-run anymore. For withdrawETH, I think the _tokenBalance change will flag stop this, because when you enter with the flashloan logic you must have pushed the funds out somehow already, but the balance has to remain the same at the end of the execution of the inner(flashloan) _bundleInternal.

Issue #392, opened 2023-03-20
SDK/frontend constant and chain refactor


ferostar:
There are a number of hardcoded values both within the SDK and the frontend that should be turned into SDK constants.

- `DEFAULT_SLIPPAGE` is the best example, but there are also a couple of URLs. We have to create a new, `commons` constants file and carefully add them. Then use them both within the SDK and the frontend.
- Chain usage is not consistent. We need to remove most of the code from `/frontend/helpers/chain` and use the SDK instead.
- This means adding a `block-explorers.ts` constants file and a corresponding property to the `Chain` class
- It also means refactor the handling of both hex string and int ids for chains. We need to always use the `ChainId` type.

PR #393, opened 2023-03-20
Chain and constants refactor

ferostar:
Implements issue https://github.com/Fujicracy/fuji-v2/issues/392.

@brozorec please have a look. 

- The chain implementation was pretty confusing. Sometimes our own `Chain` class was passed around, but others `web3-onboards`. This means we passed around chain ids as our own `ChainId`, as numbers and as hex string values in a way not very consistent. Now we only use our chains, meaning we mostly pass around `ChainId` and only convert to hex when we need to interact with the `onboard` library.
- In order to implement the above, I created a new `chain-properties.ts` constants file and convenience properties to the `Chain` class.
- I also added the `DEFAULT_SLIPPAGE` constant and a couple of URLs to a new `commons`


ferostar:
@brozorec I'm not sure about the `CONNEXT_URL` function since, well, it's not a constant 😅, but the code is much cleaner this way. Should we move it somewhere else?

packages/frontend/helpers/chains.ts
```diff
@@ -1,139 +1,84 @@
...
 
-  explorersMap.set(c.id, c.blockExplorerUrl)
-  explorersMap.set(parseInt(c.id), c.blockExplorerUrl)
+export const onboardChains: OnboardChain[] = chains.map((c) => {
+  return {
+    id: chainIdToHex(c.chainId),
+    token: c.nativeTokenName,
+    label: c.label,
+    blockExplorerUrl: c.blockExplorerUrl,
+    rpcUrl: rpcs[c.chainId],
```
brozorec, line 45: The rpc url is already available here:
```
rpcUrl: c.connection.rpcProvider.connection.url
```
so no need of the const `rpcs`

packages/sdk/src/constants/chain-properties.ts
```diff
@@ -0,0 +1,37 @@
+import { ChainId } from '../enums';
+
+export const CHAIN_LABEL: Record<ChainId, string> = {
```
brozorec, line 3: maybe "CHAIN_NAME" is better

packages/sdk/src/constants/chain-properties.ts
brozorec, line 1: I like how you did here, putting all chain props in a single place :+1:  It's worth it to move `ChainCoingeckoKey` and `ChainLlamaKey` here from "enums/ChainKey.ts" and make them Records instead of enums. Then, to remove them as parameters in the chain entity contstructor.

That will make all the chain props being managed the same way.

packages/sdk/src/constants/chain-properties.ts
```diff
@@ -0,0 +1,37 @@
...
+  [ChainId.FANTOM]: 'Fantom',
+  [ChainId.MATIC]: 'Polygon',
+  [ChainId.OPTIMISM]: 'Optimism',
+  [ChainId.GNOSIS]: 'Gnosis',
+  [ChainId.GOERLI]: 'Goerli',
+  [ChainId.OPTIMISM_GOERLI]: 'Optimism Goerli',
+  [ChainId.MATIC_MUMBAI]: 'Mumbai',
+};
+
+export const CHAIN_NATIVE_TOKEN_NAME: Record<ChainId, string> = {
```
brozorec, line 15: Those are available as:
`NATIVE[chainId].name` or `NATIVE[chainId].symbol`

brozorec (changes requested):

ferostar:
Thank you for the feedback! On it

packages/frontend/helpers/chains.ts
ferostar: `connection` can be `undefined`, I was thinking of using `??` but not sure what can we use as value without using the `rpcs` const.

packages/sdk/src/constants/chain-properties.ts
ferostar: Same as above, `NATIVE[chainId]` can be `undefined`, when `onboard` requires a value. Can I freely do `... ?? ""` since for our cases It will always have a value or should we change anything?

packages/frontend/helpers/chains.ts
brozorec: I see, I think it's safe to assume the connection is not `undefined` because we set the chain connection while initializing the sdk.
Maybe you can use the rpcs defined here as fallbacks. So when new chains get added and we forgot to update here, everything would work fine.

packages/sdk/src/entities/Chain.ts
```diff
@@ -27,6 +32,10 @@ export class Chain {
 
   connection?: ChainConnectionDetails;
 
+  label: string; // Convenience property
```
brozorec, line 36: rename to "name" to be coherent

packages/frontend/helpers/chains.ts
```diff
@@ -1,139 +1,84 @@
...
+  [ChainId.MATIC_MUMBAI]: "https://matic-mainnet.chainstacklabs.com",
+  [ChainId.OPTIMISM_GOERLI]: "https://goerli.optimism.io/",
+}
 
-  explorersMap.set(c.id, c.blockExplorerUrl)
-  explorersMap.set(parseInt(c.id), c.blockExplorerUrl)
+export const onboardChains: OnboardChain[] = chains.map((c) => {
+  return {
+    id: chainIdToHex(c.chainId),
+    token: c.nativeTokenName,
```
brozorec, line 35: `token: NATIVE[c.chainId].symbol ?? 'ETH'`

packages/sdk/src/entities/Chain.ts
```diff
@@ -27,6 +32,10 @@ export class Chain {
...
+  nativeTokenName: string; // Convenience property
```
brozorec, line 37: no longer need if change on L42 in "packages/frontend/helpers/chains.ts" is accepted

packages/sdk/src/constants/chain-properties.ts
brozorec: no longer need if change on L42 in "packages/frontend/helpers/chains.ts" is accepted

brozorec (changes requested):

brozorec (approved):

Issue #394, opened 2023-03-20
Mobile version


ferostar:
So far I tested only the app's initial state, without connecting an address and I already found three issues on a mobile device:

- Onboarding modal

<img src="https://user-images.githubusercontent.com/501915/226463249-e2a9c7e1-8f54-4ca9-93a5-5834673773af.jpg" width=50% height=50%>

- Header when there's no connected address (it's in two lines, it should all fit in one)

<img src="https://user-images.githubusercontent.com/501915/226463276-0561a1f5-a087-49d2-a167-57311e876927.jpg" width=50% height=50%>

- `@web3-onboard` modal: look at the text at the top, it's all black (it's white in the browser.

<img src="https://user-images.githubusercontent.com/501915/226463310-8952e2cc-15b8-498a-9c8b-e0b4ab5435f9.jpg" width=50% height=50%>

ferostar:
Found another one, I'm going to have a look at this one @NikolaiYurchenko :

- When there's no connected wallet, header modal shows `Walletundefined....` text.

PR #396, opened 2023-03-20
Mobile frontend

ferostar:
Implementation of issue https://github.com/Fujicracy/fuji-v2/issues/394

- [x] Undefined wallet
- [x] Onboarding modal
- [x] Two-row header
- [x] web3-onboard text

*WIP*


ferostar:
Two quick things I noticed on the `overview` dialog @NikolaiYurchenko :

- Make all top values right-aligned (they’re currently left-aligned)
- LTVProgressBar looks agreat on the desktop, but in the mobile version, the `0% ` green percentage is off, way to the left

ferostar:
I assigned this one to my self so I can't review it 😅

brozorec:
My feedback:
- onboarding modal
- borrow page: in overview, the $ value of borrow amount is always 1$
- reduce height of the second input of the borrow form and make the LTV recommendation fit on one line
image
- round up TVL %
image
- slippage settings don't appear on mobile
- menu: Wallet is stuck to address. What about Transaction history, do we have any designs for it?
image
- when I go to my mobile metamask browser and go to my-postions page, the app craches; not sure how to reproduce or inpsect this

ferostar:
> menu: Wallet is stuck to address. What about Transaction history, do we have any designs for it?

We don't have anything. When the user taps either `Wallet` or `Transaction history`, we just hide the modal. I would remove the transaction history for the time being.

ferostar:
All fixed, can't reproduce the Metamask one -probably because I tested with a wallet with no positions. Shouldn't have anything to do. Still, adding Sentry will help.

Instead of adding a new modal for the account, I simply added the one we already have. Meaning, we now have the transaction list 🥳.

Give it a try @brozorec !

brozorec (approved):

PR #397, opened 2023-03-21
Maxltv and liqratio into borrowing vault constructor

DaigaroCota:
This pull requests addresses:
- H-3
- Q-6
- Q-7

of the Macro audit report.


packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -129,16 +129,23 @@ contract BorrowingVault is BaseVault {
...
     oracle = IFujiOracle(oracle_);
-    maxLtv = 75 * 1e16;
-    liqRatio = 80 * 1e16;
+
+    if (maxLtv_ == 0 || liqRatio_ == 0 || maxLtv_ < 1e16 || maxLtv_ >= 1e18 || liqRatio_ < maxLtv_)
+    {
+      revert BaseVault__setter_invalidInput();
+    }
+    maxLtv = maxLtv_;
+    liqRatio = liqRatio_;
```
pedrovalido, line 148: This will set the maxLtv and liqRatio values to the ones being passed. Would there be a way to add an extra check regarding the providers_ passed?
Maybe (in pseudo code):


```
 function _checkMaxLtvAndLiqRatio(uint256 maxLtv_, uint256 liqRatio, ILendingProvider[] memory providers) internal pure {
     for(uint256 i; i<providers.length; i++) {
         if(providers[i].getLtv() < maxLtv || providers[i].getLiqRatio() < liqRatio)
             revert BaseVault__setter_invalidInput();
     }
 }   
```
 
 Not sure if this is possible as currently there's no getLtx ou getLiqRatio functions in the ILendingProvider.
 
 Following the current logic (without this check), it's still possible to have initial ltv and liqratio higher than the current provider if the values passed to the constructor are higher than the provider and lower than 1e18 (for maxLtv_) and 1e16 (for liqRatio_).

pedrovalido (commented):
Good job! Please take a look at the comment on the BorrowingVault constructor and let me know your thoughts

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
pedrovalido: New issue created to address this changes #438

abhishekvispute (commented):
BorrowingVault.sol LGTM 
Have only reviewed BorrowingVault.sol, since the other contract was out of scope.

Issue #264, opened 2023-01-19
Replace MockOracle with real Oracle in ForkinSetup


pedrovalido:

pedrovalido:
Note to self:
FujiOracle constructor takes `address[] assets` and `address[] priceFeeds` as an argument. This array of priceFeeds should be an `array of Chainlink contract addresses`

pedrovalido:
Note to self:
In chainlink documentation there are several addresses https://docs.chain.link/data-feeds/price-feeds/addresses#Ethereum%20Mainnet

brozorec:
New contract that ForkingSetup inherits from that should contain a mapping:

chain domain -> token addresses[]
chain domain -> feed addresses[]

or

chain domain -> token -> feed

PR #398, opened 2023-03-21
Fix Liquidation with assets under 18 decimals

pedrovalido:
This PR addresses H-2 in Macro report.

PR to fix the liquidation of assets under 18 decimals. Implemented a test inside the forking tests of liquidation manager. Inside this test, a new vault is deployed with usdc as collateral asset and weth as debt asset (usdc is under 18 decimals). The test follows the same logic as the test_liquidateDefault above implemented for the liquidation manager tests. 
The problem mentioned by macro is the liquidate function inside the borrowing vault not accounting for the assets' decimals when calculating the gained shares which led to full liquidation when the debtAsset had significantly more decimals than the collateral asset. The fix consisted on adding the decimals of collateral to this calculation:

`uint256 gainedAssets = Math.mulDiv(debtToCover, 10 ** _asset.decimals(), discountedPrice);`
`gainedShares = convertToShares(gainedAssets);`

ps.: I think the failing tests on the routers have to do with the "real oracle" branch. This was my bad on #264 and will fix it as mentioned by @DaigaroCota .


DaigaroCota:
@pedrovalido please confirm if this is ready for review. I believe as you described this branch is based on the mockOracle replacement branch right?

pedrovalido:
You're right Ill move them to the blocked until I fix the real oracle.

pedrovalido:
@DaigaroCota I've fixed the issue with the oracle and if PR has no issues, then this is ready

DaigaroCota (approved):

abhishekvispute (commented):
lgtm

PR #406, opened 2023-03-23
Position yields modal

NikolaiYurchenko:


NikolaiYurchenko:
> A couple of questions @brozorec
> 
> * The elements in the list appear clickable, but nothing happens. Should we remove the pointer or take the user to the position screen?
> * At the top we have `estEarnings >= 0 ? 'Earnings' : 'Costs'`, but as a list header we have just `earnings`, what should we do there?
> 
> @NikolaiYurchenko can you have a look at the position screen in a mobile device? Now that we added the "view" button, next to the "borrow more", doesn't look too good. If you have a look at the [design](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=3095-53556&t=wqbObV7nTHOaNZEx-4), the button should be in a third row.

@ferostar thanks for the review, I'll fix the adaptations.
Regarding the header title, maybe smth like Est. USD Value Change (todo: think about it)

ferostar:
> @ferostar thanks for the review, I'll fix the adaptations. Regarding the header title, maybe smth like Est. USD Value Change (todo: think about it)

Yes, I wasn't sure about it. I considered a tooltip, since we don't have any in that modal, since it's a table header and it shouldn't be long, but just not sure 🤔.

ferostar (approved):

brozorec (approved):

PR #408, opened 2023-03-24
Balance polling

ferostar:
Ok, so I implemented a basic polling mechanism. Some notes, both to test and feedback:

- It happens at the `_app.tsx` level.
- If the user is borrowing or managing a position, it refreshes the `ERC20` tokens with our `sdk`. 
- Otherwise, we only trigger a native token balance update with `@web3-onboard`.
- Updating happens every 15 seconds, but I added a `constant` we can quickly change.
- Since a lot of things can go south, I added a number of checks to the collateral/debt updates. Namely: there can't be any actions going on (if the store variables `isExecuting` or `isSigning` are true, we don't do anything) and before updating the values we make an extra check to see if the chains haven't changed.
- I took the opportunity to change `constants` to match the `sdk` a bit more.
- I added a little thing and now, once the user makes a tx, in history store's `watch` function, as soon as we have a receipt I check if the tx happened on main chain and trigger the same native token balance update -that way, if there was a noticeable gas fee, the user will see it pretty quickly.


packages/frontend/helpers/balances.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  if (!shouldFetchERC20) return;
+
+  // Now let's to with ERC20 tokens
+  const debt = useBorrow.getState().debt;
+  const collateral = useBorrow.getState().collateral;
+
+  const data = [collateral, debt];
+
+  data.forEach(async (asset, i) => {
```
brozorec, line 48: I'd suggest creating a new function `checkBalance(type)` that will be called from the current `checkBalances()` so that we'll avoid the forEach for just two elements and referencing "types" by index.

packages/frontend/pages/_app.tsx
```diff
@@ -46,13 +51,26 @@ function MyApp({ Component, pageProps }: AppProps) {
...
+  }, [address]);
+
   useEffect(() => {
     const handleRouteChange = (url: string) => {
       const isTop = isTopLevelUrl(url);
       if (isTop && address) {
         fetchPositions();
         updateVault();
       }
+      const should = url === '/borrow' || url.includes('/my-positions/');
```
brozorec, line 72: Isn't it better to have all urls as constants in a separate file so that if we want to change them, to have them in a centralized manner?

brozorec (dismissed):

brozorec (approved):

PR #411, opened 2023-03-27
Toastify implementation

NikolaiYurchenko:
Added react-toastify package and implemented it in 3 ways:
1. Success notification with transaction URL (replacement of current)
2. Basic 4 types of transactions 'error' | 'info' | 'success' | 'warning'
3. same types but with link.

Also removed snackbar and changed button label from 'Need to borrow at least 1 USD' to 'Need to borrow at least 1 USD'
because we have condition like if debtAmount <= 1

Also, didnt create state with list of notificarion because didn't find a usecase where we want to store them on FE side


ferostar:
Hey @NikolaiYurchenko !

I can't even run the project after installing `react-toastify`, I get `Module not found: Can't resolve 'react-toastify/dist/ReactToastify.min.css'` 🤔.

ferostar:
You're right @NikolaiYurchenko there's no need for a store, just get rid of it and move the `notify` function to the `helper`.

brozorec (approved):

PR #412, opened 2023-03-27
Internal v2 testing

ferostar:
- [x]  Manage positon: add tooltip on why we can’t change borrow or payback chain
- [x]  Manage postion: update vault balances after operation
- [x]  Overview: borrowed $ value is hard-coded (always 1$)
- [x]  Fuji Finance in footer
- [ ]  ~~Wording “Add Position” and “Remove Position” // Wording 🤔?~~
- [ ]  ~~Connext explorer link appears too early (Dagiaro) // Research~~
- [x]  Auth issues when wallet but no account
- [x]  Remove current ltv in borrowing
- [x]  Current ltv changes, should be the base position value, not futurePosition
- [x]  Remove social icons from main screen
- [x]  Remove approval modal, take user straight to metamask 💪
- [x]  $ sign in debt?
- [x]  default to the user’s chain if supported
- [x]  Change "above" to "below" and define color according to recommended ltv
- [x]  Tx list shows check for all not-ongoing txs, even failed ones
- [x]  Manage position, remove position: withdraw max
- [x]  Borrow and manage-position adding position, recommended ltv tappable
- [x]  Loading spinner never disappears
- [ ]  ~~Add contextualized placeholders in inputs, for example instead of displaying 0 when withdrawing, display “Type amount to withdraw” // Question~~
- [x] Tx modal: show destination icon instead of connext and destination link for bridge steps
- [x] Account modal: show last 10
- [x] Take care of dust in both my positions and borrow screen // Started, need input
- [x] Max button removing position calculations
- [x] Borrow limit incorrectly calculated
- [x] Replace "i" with hover tooltip and add text
- [x] Show "approving" as the button title while its beeing processed


brozorec:
Some feedback on the changes:

1. On "Remove position": 
The "max" button should work differently; it shouldn't fill the input with the user's balance. Rather, we have to calculate what's the max amount the user can withdraw from this position:

```
MAX_COLLATERAL_WITHDRAW = 
    COLLATERAL_BALANCE - 
    (BORROW_BALANCE - PAYBACK_INPUT) / (MAX_LTV * COLLATERAL_PRICE)
```

2. It seems that the "Borrow Limit" isn't calculated the right way, maybe because it doesn't take into account some of the inputs

```
BORROW_LIMIT = 
  (COLLATERAL_BALANCE + COLLATERAL_DELTA) * COLLATERAL_PRICE * MAX_LTV - 
  (BORROW_BALANCE + BORROW_DELTA)
```
COLLATERAL_DELTA = - COLATERAL_INPUT if we withdraw
COLLATERAL_DELTA = + COLATERAL_INPUT if we deposit
similar for BORROW_DELTA

3. On manage-position, we've put "i" with a tooltip next to the disabled chain. Can we put just the tooltip on hovering over the disabled dropdown?
Here's the text for the tooltip:
- "Paying back an existing position from another chain will be enabled soon"
- "Borrowing from an existing position to another chain will be enabled soon"

4. When approving spending limit, change the button from "Approve" to "Approving ..." when the tx is pending.

ferostar:
@NikolaiYurchenko a little something about the recommended feature

- If there is no collateral, the button doesn't do anything. It should, since the user should be able to simply borrow more.
- The new button isn't properly centered

ferostar:
@brozorec some stuff we need to continue with some of the items:

- Need words for `Add position` and `remove position`.
- Need placeholder texts.
- Need to know dust value (I'm currently checking both debt and collateral the same way, but given that one is WETH and the other is USD I'm guessing we need to do It in different ways?)
- What do we do with the "connext explorer link appears too early" issue?
- Can you clarify the borrow limit calculation?

brozorec:
1. Tx History
- [ ] when only one action payback or deposit, there's a trailing "and"
image
- [ ] use icons according to the design; if they are not available in Figma, ask Mark

2. Approval process
- [ ] the height of the approval button changes when the state is pending and there's no text "Approving" (only on borrow page)
- [ ] when approval is successful, display a notification with a link to the chain explorer

3. Borrow limit calculation
After I tested it out again, I figured out that we shouldn't take into account the borrow balance and the borrow input. So here's how it should work. The calculation depends on whether we are adding position or removing position. In the first case, we have to add the collateral input, while in the second, we have to deduct it:

- when on add position
```
BORROW_LIMIT = 
  (COLLATERAL_BALANCE + COLLATERAL_INPUT) * COLLATERAL_PRICE * MAX_LTV
```
- when on remove position
```
BORROW_LIMIT = 
  (COLLATERAL_BALANCE - COLLATERAL_INPUT) * COLLATERAL_PRICE * MAX_LTV
```

5. Recommended LTV in input
- [ ] keep how it was looking before (Recommended 55%); just make it clickable with a pointer on hovering

6. The max button when withdrawing doesn't work correctly, although I see the formula was implemented right.
When
COLLATERAL_BALANCE = 0.001
BORROW_BALANCE = 5.31
BORROW_INPUT = 0
COLLATERAL_PRICE = 1799
MAX_LTV = 75%
then
MAX = 0.001 - (5.31 - 0) / (0.75 * 1799) = 0.001066667
while I have
image

Could the issue be that `maxLtv` isn't taken as a %?

7. Let's consider dust if the value is lower than 0.0001
8. Regarding the input placeholders and the titles for "Add Position" and "Remove Position", let me sync with @Markoyw 
9. For the issue with the Connext scan, Idk yet

brozorec:
Updates about some of the unresolved points in the first comment:

1. Wording “Add Position” and “Remove Position”
2. Add contextualized placeholders in inputs
@Markoyw is going to propose new designs so I suggest we do them in a separate PR.
---
3. Connext explorer link appears too early
This will also be handled in another PR that will address #423
---
4. Take care of dust in both my positions and borrow screen
We set an amount for dust to 0.0001 and we check **only** the collateral.
---

@NikolaiYurchenko Can you go through the feedback of @Markoyw from [here](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=5026%3A72064&t=Wisejam29Oon0ij4-1) and check if the comments about alignments, modals, font sizes and the other pure UI stuff are already handled?

brozorec:
- [x] I created a ticket for Mark's feedback https://github.com/Fujicracy/fuji-v2/issues/430
- [ ]  There's an issue with the borrow limit 
image
- [ ] When I click on the "Recommenened LTV", it gets to 35% and not 55%
image
- [ ] Change "Connext" to "Polygon network"
image
- [ ] I selected "max" withdrawal and clicked "Withdraw", but nothing happened. I got an error in the console `borrow.store.ts:838 Uncaught (in promise) Unexpected undefined param`.
image
When I delete the last digit of the amount, it works
image
- [ ] When it's a single operation, it starts with "and" 
image
- [ ] Text below steps should be centered 
image

PR #416, opened 2023-03-28
Sdk error handling

ferostar:


packages/sdk/src/types/FujiResult.ts
```diff
@@ -0,0 +1,5 @@
+import { FujiError } from '../entities/FujiError';
+
+export type FujiResult<T> =
+  | { success: true; data: T }
+  | { success: false; error: FujiError };
```
brozorec, line 5: What about creating 2 entities: `FujiResultError` and `FujiResultSuccess` (similar to `Token` and `NativeCurrency` that are both `Currency`):
```
export type FujiResult<T> = FujiResultSuccess<T> | FujiResultError;
```

That will allow to return `new FujiResultError(...)` and `new FujiResultSuccess(...)` instead of 

```
return {
  success: true,
  data: ....
}
```

packages/sdk/src/Sdk.ts
```diff
@@ -213,27 +216,24 @@ export class Sdk {
...
-    for (const chain of chains) {
-      const chainId = chain.chainId;
-      const vaults = VAULT_LIST[chainId].map((v) =>
-        v.setConnection(this._configParams)
-      );
-      const v = await batchLoad(vaults, account, chain);
-      res.push(...v);
+    chainId: ChainId,
+    account?: Address
+  ): Promise<FujiResult<VaultWithFinancials[]>> {
```
brozorec, line 225: Can we have a wrapper `FujiResultPromise<>`?

ferostar: After creating the wrapper `export type FujiResultPromise<T> = Promise<FujiResult<T>>`, I can't use it from the frontend since I get the following error

```Type 'FujiResultPromise' is not a valid async function return type in ES5/ES3 because it does not refer to a Promise-compatible constructor value.ts(1055)```

It works when calling the `sdk`, but we sometimes want to return the same type from the app, and that fails with the above error. Thoughts @brozorec ?

File: packages/sdk/src/types/FujiResult.ts (hunk 1 above)
Comment (ferostar) [reply packages/sdk/src/types/FujiResult.ts:5]: Initially, I thought it was a great idea but after quickly implementing it started having second thoughts.

First, it forces us to start typecasting or heavily extending. If we don't extend from the same class, we'll end up doing

```
if (result instanceof FujiResultSuccess) { 
  ... cast
}
```

Otherwise, we'll have to create yet another `FujiResultParent` class and have both `...Success` and `...Error` extend from it, then it would be a matter of checking for its properties. I think having a type with props would be more readable and way more standard.

packages/sdk/src/Sdk.ts
brozorec: Found [this](https://github.com/microsoft/TypeScript/issues/26781). Can you try changing target in tsconfig to "es6", tho Idk what other consequences it can have.

packages/sdk/src/types/FujiResult.ts
brozorec: Wouldn't be enough just to check `if (result.success)`? And then inside being able to access `result.data` or `result.error` depending on the condition?

ferostar: You're right, I didn't realize it was a matter of adding the `success` property to each just like with the `type`.

packages/sdk/src/Sdk.ts
ferostar: That was it. I thought everything would break, that's why I didn't even consider changing the target, but did it after fixing solving the other comment and it worked smoothly without any issues (so far), so I'll push.

ferostar:
- Changed the error and its result constructors to take the code as the second param, which is now optional for the result. The default value is our `SDK` error code, so most calls are simpler. Plus… message is passed to super, so makes more sense to have it as the first one.

- Also changed all of the Preview functions and their implementation in the app

- Applied the changes to `validateAndParseAddress` but reverted them. It’s used in `Address.ts`’s constructor, and I can’t realize why yet, but the frontend starts failing. Not sure if we want to follow the same approach for all entities, so have a look if you want to @brozorec .

PR #424, opened 2023-04-03
Connext tx details

brozorec:
@ferostar @NikolaiYurchenko I propose in this PR a new method `sdk.getConnextTxDetails`. The point is to stop using `sdk.watchTxStatus` for tracking all kind of transactions but to split the flow in 2, one flow for single-chain txns and one for cross-chain txns.


ferostar:
> @ferostar @NikolaiYurchenko I propose in this PR a new method `sdk.getConnextTxDetails`. The point is to stop using `sdk.watchTxStatus` for tracking all kind of transactions but to split the flow in 2, one flow for single-chain txns and one for cross-chain txns.

And we'll poll each couple of seconds and update the UI until having results, right? Makes sense @brozorec . I'm making a couple of small but breaking changes, going to update the branch as soon as I finish with them.

ferostar:
@brozorec PR is already up to date with `sdk/feat/error-handling`, should we merge it?

PR #429, opened 2023-04-04
Warning block on existing position

NikolaiYurchenko:
Warning block on Borrow page to show in same open position exists

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -33,7 +33,6 @@
     "debounce": "^1.2.1",
     "ethers": "^5.7.0",
     "immer": "^9.0.16",
-    "mixpanel-browser": "^2.45.0",
     "next": "^13.2.1",
     "react": "18.2.0",
     "react-dom": "18.2.0",
@@ -44,7 +43,6 @@
   "devDependencies": {
     "@synthetixio/synpress": "^2.4.1-beta.1",
     "@types/debounce": "^1.2.1",
-    "@types/mixpanel-browser": "^2.38.0",
     "@types/node": "18.7.13",
     "@types/react": "18.0.17",
     "@types/react-dom": "18.0.6",
--- a/packages/frontend/pages/_app.tsx
+++ b/packages/frontend/pages/_app.tsx
@@ -2,7 +2,6 @@ import '../styles/globals.css';
 
 import { ThemeProvider } from '@mui/material';
 import { Web3OnboardProvider } from '@web3-onboard/react';
-import mixpanel from 'mixpanel-browser';
 import { AppProps } from 'next/app';
 import { Inter } from 'next/font/google';
 import { useRouter } from 'next/router';
@@ -36,9 +35,6 @@ function MyApp({ Component, pageProps }: AppProps) {
   const fetchPositions = usePositions((state) => state.fetchUserPositions);
 
   useEffect(() => {
-    mixpanel.init('030ddddf19623797be516b634956d108', {
-      debug: process.env.NEXT_PUBLIC_APP_ENV === 'development',
-    });
     initErrorReporting();
     initAuth();
   }, [initAuth]);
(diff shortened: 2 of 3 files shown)


packages/frontend/components/Borrow/Wrapper.tsx
```diff
@@ -100,14 +100,10 @@ function BorrowWrapper({ query }: BorrowWrapperProps) {
   return (
     <>
       <Head>
-        <title>{`${isEditing ? 'Position' : 'Borrow'} - xFuji`}</title>
+        <title>Position - Fuji V2 Himalaya</title>
```
ferostar, line 103: With this change, the browser will show "Position" as the title instead of "Borrow" when the user creates a new position, which is super confusing. We're trying to distinguish between them as much as possible. I think we need to change it @brozorec .

ferostar:
@brozorec the `previews` functions fail if we send `0` as value.

ferostar:
- [ ] Updating the `sdk` functions so they work sending `0` or adding a new one so it's possible to fetch the routes before the users enters any values

Then...

- [ ] Remove the `failureForMode` check since the idea is to always trigger the call (if there's an active vault)
- [ ] Make sure we trigger an update to the `meta` as soon as we have an active vault
- [ ] Prevent showing the warning sign if we don't have route information
- [ ] Update the warning text: it's not just a matter of changing a chain but changing to a different route

Made the PR into a draft again since there's still a lot to do.

ferostar:
@NikolaiYurchenko turned the PR into a draft again and added a checklist.

packages/frontend/components/Borrow/Wrapper.tsx
brozorec: That's true, agree to revert

ferostar:
@brozorec when calling `preview` calls in a cross-chain operation and we pass 0 in both amount fields, we get the following error from the `sdk`:

```
Sdk error while attempting to set meta: division-by-zero [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-division-by-zero ] (fault="division-by-zero", operation="div", code=NUMERIC_FAULT, version=bignumber/5.7.0)
```

ferostar:
@NikolaiYurchenko there was a leftover UI task, routing card icons aren't centered, pls fix fix it here (and make sure to improve the code while doing it, there's a lot of code duplication in that component when rendering the steps)

Captura de pantalla 2023-04-06 a la(s) 17 12 06

brozorec:
@ferostar @NikolaiYurchenko great job on this and the following PR :muscle: 

I have these suggestions to improve the warning messages and the routing modal:

### Messages
1. when there's a single route

"Based on your selection, we've noticed that you have an open WETH/USDC position on (icon) Optimism. You may proceed to manage it."
- _Replace WETH, USDC and Optimisim with the corresponding values._

2. when there's more than one route

"Based on your selection, we've noticed that you have an open WETH/USDC position on (icon) Optimism. You may proceed to manage it. 

If you're trying to open a similar position on another chain, please [select a different route]."

- _Replace WETH, USDC and Optimisim with the corresponding values._
- _The last sentence is on a new line._
- _[select a different route] is underlined and clickable and opens the routing modal._

### Routing modal
Regarding the routing modal, displaying zeros may be confusing, so I suggest hiding them:

- when routes are empty
image
We may display the following:
1. WETH from Optimism
2. Bridge to Gnosis via Connext
3. Deposit WETH
4. Borrow USDC
5. Get USDC on Gnosis

- when routes are populated
image
We may display the following:
1. 0.13 WETH from Optimism
2. Bridge to Gnosis via Connext
3. Deposit 0.13 WETH
4. Borrow 12 USDC
6. Get 12 USDC on Gnosis

ferostar:
Another important issue @brozorec is that we just don't show the routing modal on a mobile device.

ferostar:
Turned the comments above into checklists so it's clear what remains to be done

NikolaiYurchenko:
Message changed, on mobile underline is hidded, returned input check becase Mark told me to do it

ferostar:
Have a look @brozorec !

brozorec:
I removed the chain icon because it didn't look good (sorry about that). I also used the vault values because a position is determined by the vault where it was open, not by the start and the end step. The steps define only the route.

ferostar (approved):

PR #434, opened 2023-04-04
Transaction confirmation modal

NikolaiYurchenko:
Transaction corfirmation flow rework;
LtvWarning modal  removed
Modal pops after approval if needed


ferostar:
- [ ] In the route cards, the line is bigger when there are icons, so it looks a bit poor
Captura de pantalla 2023-04-05 a la(s) 15 37 20

- [ ] `Borrow limit left` is always 0 when creating a new position. Did an operation with an `LTV` of around `30%`, so it shouldn't be 0 at all. Also, there's already a `borrowLimit` function in the `assets` helper. While it has a bug atm, you can use it here. In this confirmation modal, the idea is to show how much could the user borrow. The "...left" part. So if the user is doing an operation with an LTV of 30%, it means they could borrow a lot more until getting the recommended value. That's what we want to show.

ferostar:
@NikolaiYurchenko you're still using a different function to calculate the borrowing limit and doing it twice. Please reuse the `assets` function as mentioned earlier, pass different params in order to get different results and apply the changes to the result in order to get the calculation. We're duplicating code without need at the moment, this is going to be an issue in the future (btw, there shouldn't be any bugs left in the current implementation, make sure to grab [internal-v2-testing](https://github.com/Fujicracy/fuji-v2/tree/frontend/fix/internal-v2-testing)'s latest.

ferostar:
Routes still have varying line heights depending if there's an icon or not.

Captura de pantalla 2023-04-06 a la(s) 11 48 26

ferostar:
- [ ] no scrolling
- [ ] hide 'via connext' in single chain txs

ferostar:
@brozorec @NikolaiYurchenko 

1- Simplified the `max borrow limit` function. Before it was more clear, but as Nicko said, there was no need to do a calculation since we already had the value to use!
2- The `borrow limit left` function now calls the max borrow limit and works on its result. Also changed the params.
3- Renamed both so it's more clear what it is they actually do.

packages/frontend/components/Shared/ConfirmTransactionModal.tsx
```diff
@@ -131,7 +131,12 @@ export function ConfirmTransactionModal({
           <></>
         )}
 
-        {steps && steps.length > 0 && <RouteBox steps={steps} />}
+        {steps && steps.length > 0 && (
+          <RouteBox
+            steps={steps}
+            isCrossChainTransaction={collateral.chainId !== debt.chainId}
```
brozorec, line 160: This is not a precise way to check if a tx is a cross-chain one because not always we have both collateral and debt (We can only have a deposit action and do this from another chain which makes it a cross-chain tx).

You can check if a tx is cross-chain by comparing `chainId` of the START step and the END step.
```
const start = steps.find(RoutingStep.START)
const end  = steps.find(RoutingStep.END)
const isCrossChain = start.chainId !== end.chainId
```

brozorec (changes requested):
image
- [x] In the upper section, start with "Payback", then "Withdraw"
- [x] When payback, change "from" to "to" (Payback USDC to Agave)

---
- [ ] @ferostar @NikolaiYurchenko Yesterday, we missed something important when we talked about the upper action elements.

We were basing our arguments on this implementation:
image

While it was different than the proposed designs:
image

The differences are that in the 1st image, the network is the one from which the collateral comes from, and not the one where it's supposed to be deposited. The second is that the explanation message should differ depending on the action. 

If we take into account that, I think the designs make sense. Though, we have to shorten a bit the explainers so that they fit on one line.

- [x] Here are my propositions for every action:
- Deposit collateral: "The network where the collateral will be deposited on"
- Borrow: "The network where the loan will be taken out"
- Payback: "The network where the debt will be paid back"
- Withdraw: "The network from where the collateral will be withdrawn"

There's also a popover, but it's talking about routes, and it's only valid in the context of an initial borrow and not when managing an existing position. (we don't have routes on the manage position page). Is it possible to display it according to the context? If yes, here's its content should be:
"If you wish to open a position on a different chain, please select an alternative route."

packages/frontend/components/Shared/ConfirmTransactionModal.tsx
```diff
@@ -0,0 +1,301 @@
...
+        <InfoRow
+          title="Est.processing time"
+          value={
+            <Typography variant="small">{`~${
+              transactionMeta.estimateTime / 60
+            } minutes`}</Typography>
+          }
+        />
+
+        {collateral.chainId !== debt.chainId && (
```
ferostar, line 166: This check is also poor @NikolaiYurchenko , need to change it according to Boyan's comment.

brozorec (changes requested):
With this 
image
we are in the same place as yesterday.

Depositing and borrowing should always be on the same chain, same for payback and withdraw.

I think the `AssetBox` is more complicated than needed, leading to wrong results. For example, here, this is supposed to be a cross-chain tx. I start from Gnosis and the collateral should be deposited to the position on Optimism:
image

However, the confirmation modal doesn't display it:
image

I suggest refactoring the AssetBox so that it takes a `step`. The `step` contains everything we need: `chainId`, `token`, `amount` and the name of the step (Deposit, Borrow, ...). We additionally have to pass only `isEditing` that's used for the popover. 

In "ConfirmationModal", instead of having two `AssetBox`, iterate through the array `steps` and filter only "Deposit", "Borrow", "Payback" and "Withdraw" steps.

ferostar:
@NikolaiYurchenko 

Passing

```ts
<BorrowButton
 ...
  withConfirmation={proceedWithConfirmation}
 />
```

results in

```
Type '{ address: string | undefined; collateral: AssetChange; debt: AssetChange; position: Position; walletChain: ConnectedChain | undefined; ... 15 more ...; withConfirmation: (action: () => void) => void; }' is not assignable to type 'IntrinsicAttributes & BorrowButtonProps'.
  Property 'withConfirmation' does not exist on type 'IntrinsicAttributes & BorrowButtonProps'.ts(2322)
```

NikolaiYurchenko:
@brozorec made changes we talked about. Please check
You were right, if we dont need those helpers right now, code became a little smaller

brozorec (approved):

PR #439, opened 2023-04-06
Show error notifications

ferostar:


brozorec:
1. When I disconnected manually from MM, I got this error thrown
image
2. I'm disconnected, I go to the markets page and select a vault on Arbitrum. I'm redirected to the borrow page, I click "Connect Wallet" and got this. In the meantime, the deposit and the borrow chains were changed to Polygon.
image
3. I'm connected with an address with an existing position on Arbitrum, my wallet chain is Arbitrum and I start from the borrow page. Getting again the same error (about the 3-chain transfers) and don't have the warning message for the existing position.
4. Happening the same when I click on the max button in the collateral input.

ferostar:
> I tested this a lot a couple of weeks ago -it never crashed. I'm guessing the latest MM upgrade broke things.

Turns out it is `@web3-onboard/core` the one that's crashing. When the wallet is disconnected, a subscription is triggered and the following line crashes because wallets is an empty array `[]`.

`const { accounts } = wallets.find(wallet => wallet.label === label);`

brozorec (approved):

PR #442, opened 2023-04-11
Do not change vault when editing

ferostar:


ferostar:
@brozorec have a look, it was a one-liner.

Issue #430, opened 2023-04-04
UI fixes


brozorec:
Based on @Markoyw feedback, here are some bug fixes, mainly on the UI side. 

- For more information, check the comments in Figma on each screen.
- Start a new branch from "frontend/fix/internal-v2-testing" if it's not already merged in "main".

- [x] align icons of providers: [link](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=5026%3A72067&t=E3cUvlIpGzPah8FS-1)
- [x] edit tx modal according designs: [link](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=5067%3A103122&t=B71VEurTgR6vhpYE-1) and [link](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=5067%3A139450&t=B71VEurTgR6vhpYE-1)
- [x] edit tx modal: close button (+ handle mobile clipping)
- [x] edit tx modal: image sizes
- [x] edit tx modal: text overlapping (works with aave, compound not)
- [x] edit tx modal: add token bottom margin too big
- [x] edit tx modal: add token color
- [x] in borrow's button comp, don't show "withdrawing" etc but "processing" for all
- [x] set fixed height background in my-positions: [link](https://www.figma.com/file/LkNVe5ob1LjGpbGU74Hcrs/Fuji---Design-Files-(DUP)?node-id=5026%3A72064&t=B71VEurTgR6vhpYE-1)

ferostar:
What is the `Validate Transaction` step in the transaction modal @brozorec ? Because we just changed the bridge step to show destination data since everyone was confused about it.

brozorec:
> What is the `Validate Transaction` step in the transaction modal @brozorec ? Because we just changed the bridge step to show destination data since everyone was confused about it.

It's just wording, we can ignore it

PR #447, opened 2023-04-14
UI fixes #2

### LINKED ISSUE #2: Front-end Architecture
Investigations:
- [x] Determine library for web3 connection
- [x] Determine app state -- How different components talked to each other, how do we maintain of what the user is doing so in the next page it is available for example
- [x] State of the transactions, how do we deal with this topic? In single blockchain we have several states of a transaction and going cross chain will require tracking those states on different blockchains.
- [x] Using GraphQL for querying Connext transaction status, mainly learn how it works in case it is needed.
- [x] How to query states from multiple chains at the same time, not relying only on the injected provider.
- [x] Typescript

---
**Todo**

Create an app in frontend with
- [x] Next.JS & TypeScript,
- [x] #37
- [x] Add xstate
- [x] #39
- [x] Add Onboard + basic onboarding and put user in react context

NikolaiYurchenko:
close https://github.com/Fujicracy/fuji-v2/issues/430
Screenshot 2023-04-14 at 14 04 52
Screenshot 2023-04-14 at 13 37 32


brozorec (changes requested):
Good job :muscle: I'd only ask for a small change in the buttons section:
image
- switch the places of "View transaction on ConnextScan" and "Add WETH"
- make all button heights equal
- adjust margins between buttons
- "Add WETH" is the least important so it can be a ghost button
- "View transaction on ConnextScan" is secondary, so it can be an outlined button


image

NikolaiYurchenko:
@brozorec changed order, made buttons medium, gap between them is 0.5, View transaction on ConnextScan variant made not outlined but secondary

brozorec (approved):

PR #458, opened 2023-04-14
Balance updating issues

ferostar:
I noticed there were problems with balances not being properly updated in the feedback reports and so I had a quick look at the code, noticing a couple of issues.

- Updating after an operation was unnecessarily complex -I simplified it and loosened some of the checks.
- Noticed that polling was broken depending on where the user first landed it.

### DIFF
--- a/packages/frontend/helpers/history.ts
+++ b/packages/frontend/helpers/history.ts
@@ -1,9 +1,7 @@
 import { ChainId, RoutingStep, RoutingStepDetails } from '@x-fuji/sdk';
 
-import { useAuth } from '../store/auth.store';
 import { useBorrow } from '../store/borrow.store';
 import { updateNativeBalance } from './balances';
-import { hexToChainId } from './chains';
 
 export enum HistoryEntryStatus {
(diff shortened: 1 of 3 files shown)


brozorec (changes requested):
I think on a cross-chain transaction, it doesn't work:

I manage my position on Polygon and I want to payback and withdraw some WETH to Gnosis. Here's the screen before submitting the transaction, my WETH balance on Gnosis is 0.00279:
image

After the transaction gets confirmed on Gnosis, my WETH balance didn't change:
image

ferostar:
> I think on a cross-chain transaction, it doesn't work:
> 
> I manage my position on Polygon and I want to payback and withdraw some WETH to Gnosis. Here's the screen before submitting the transaction, my WETH balance on Gnosis is 0.00279: image
> 
> After the transaction gets confirmed on Gnosis, my WETH balance didn't change: image

There was a simple bug with logical operators, no wonder it didn't work for you.

brozorec (approved):

PR #470, opened 2023-04-18
Borrowing issues

### LINKED ISSUE #469: Borrowing issues
- [x] When entering the borrowing screen directly, vault isn't updated
- [x] When entering the borrowing screen through markets with no account, once the user connects it overrides chains
- [x] When calls to updateVault fails the user is still able to use the borrowing component. If `availableVaultsStatus` is `error` we need to disable the button and show a notification
- [ ] Take care of issues with default chains

ferostabio:
close #469

### DIFF
--- a/packages/frontend/components/Borrow/Borrow.tsx
+++ b/packages/frontend/components/Borrow/Borrow.tsx
@@ -12,12 +12,11 @@ import { Address } from '@x-fuji/sdk';
 import { useRouter } from 'next/router';
 import React, { useEffect, useMemo, useState } from 'react';
 
-import { PATH } from '../../constants';
 import { DUST_AMOUNT_IN_WEI } from '../../constants';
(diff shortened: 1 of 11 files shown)


ferostar:
Made a couple of big changes @brozorec .

- We were doing a couple of unnecessary when the screen was loaded
- We were showing both the warning and "Manage position" while fetching the data or even if there was an error
- There was no check after doing asynchronous calls to make sure nothing had changed. Let me give you an example

```ts
const token = get().debt.token
const result = await someRandomAsyncTask(token)
// Here we had no checks. If the user changed token/chain/anything while doing the async task, 
// we still updated the value in the store
set({ result.data })
```

This is why if the default chain was Polygon but we already showed Gnosis, we could even see the WETH max balance of Polygon, since by the time we finished fetching them chain had changed to Gnosis but we set the value anyways.

PR #471, opened 2023-04-18
Transaction history

ferostar:
Until now the transaction history was deleted every time the user manually disconnected his wallet and nothing happened if he changed wallets -we showed all txs, no matter the address.

The transaction history was just an array of hashes, nothing to do.

I refactored both the transaction and entry history to include the user’s address. Now we don’t delete anything and only show the current address’s recent history both in the list and in pending, plus show notifications only for the current address.

I also had to also move the watching out of the rehydration since we want to trigger when the user changes account and I had to make sure not to watch any transactions twice.


ferostar:
Give this PR a look @brozorec !

Important notice: this work is a breaking change -while it clears the history store if it encounters string hashes during rehydration, going back to the old code after using it is a big no. If you do want to use any of the old code after making transactions with this one, you'll have to return `return <></>;` in `AccountModal`.

brozorec (approved):

Issue #468, opened 2023-04-18
UI issues


ferostar:
- [x] Inter font should show in all browsers
- [x] Inter font should show with no delay
- [x] Double check that all of the right fields use the Inter font
- [x] Mobile header is shown for iPads
- [x] Safety rating is just "..." in iPads
---
- [x] Collateral asset name gets cut off on bigger screens (displaying WE... instead of WETH)
![Screenshot from 2023-04-19 11-11-38](https://user-images.githubusercontent.com/9572072/233044792-980035e8-140d-46c3-9d82-cd25514efccd.png)
- [x] Safety rating also gets cut off in some cases
![image](https://user-images.githubusercontent.com/9572072/233045169-69417cc2-07bf-4e6a-919d-029bfde63791.png)
- [x] Display "Why do I need to sign?" only when the action includes "sign". We need to hide it when the user needs first to do an approval.
![image](https://user-images.githubusercontent.com/9572072/233046396-0c8321d5-eba2-4a70-a06c-98d846d367b4.png)

brozorec:
@NikolaiYurchenko FIY I added three more items to this ticket (there are two items for the Safety rating that are probably linked)

PR #472, opened 2023-04-19
Tablet UI fixes

NikolaiYurchenko:
closes https://github.com/Fujicracy/fuji-v2/issues/468


NikolaiYurchenko:
@ferostar @brozorec Also changed font style to use local font files again because it is faster

NikolaiYurchenko:
One more point: link tag in head of html still here and next downloads font secont time. Cant figure where it happens yet, worth taking a look later

ferostar:
Just in case @NikolaiYurchenko are all items on the issue done?

NikolaiYurchenko:
@ferostar Yes, it should be

ferostar (approved):

PR #481, opened 2023-04-20
Onboarding wallets

### LINKED ISSUE #474: Migrate to Wallet Connect v2
`@web3-onboard` upgraded their sdk to support Wallet Connect v2. Now when setting things up we need to specify both v1 and v2 arguments, and v2 requires an API key.

ferostabio:
close #474


ferostabio:
Added `coinbase` and `ledger` @brozorec .

Trezor requires registration, we need to pass at least `email` and `appUrl`:

```ts
type TrezorOptions = {
  email: string
  appUrl: string
  customNetwork?: CustomNetwork
  filter?: Platform[]
  containerElement?: string
}
```

I was looking at the list and I would also add `MyEtherWallet` and `Trust Wallet` (despite the recent security issues, it has a lot of users and they would be able to connect directly instead of using WalletConnect)

brozorec:
support@fuji.finance
https://app.fuji.finance

ferostabio:
Installed `trezor` plus a couple more wallets. Two notes.

- Didn't test the trezor wallet.
- I misread, there's no module for `Trust Wallet` but it is detected if the browser extension is installed.

brozorec (approved):

Issue #480, opened 2023-04-20
UI issues


ferostabio:
- [x] A call failed when I arrived to the borrow screen (had connectivity issues) and the token card showed `NaN` in 

```ts
<Typography variant="small" sx={{ width: '11rem' }}>
   {formatValue(usdPrice * +value, { style: 'currency' })}
</Typography>
```

We should add a `isNan` check and return a properly formated `0` instead.

- [x] When we add a token to Metamask, the UI is... different from what we now have 😅.

<img width="467" alt="Captura de pantalla 2023-04-20 a la(s) 16 37 12" src="https://user-images.githubusercontent.com/501915/233473184-fb838fb2-3f36-458f-b895-fbc8c8bf1d4e.png">

At least we should replace the emoji with our checkmark svg?

- [x] Custom slippage doesn't fit on screen

<img width="337" alt="Captura de pantalla 2023-04-20 a la(s) 18 25 44" src="https://user-images.githubusercontent.com/501915/233493902-f2a3cbaf-a3c6-4a66-9499-635545f92531.png">

- [x] Remove the circle shadow of checkboxes in the disclaimer modal
- [x] Display the amount instead of a % in the borrow input as recommended

NikolaiYurchenko:
@ferostabio I looked for what you mean in 4 point and didn't find it, can you please provide more info about this?

NikolaiYurchenko:
@ferostabio point 2 is outdated, doublechecked this with Mark. 
P.S. marked checkbox as done for this

NikolaiYurchenko:
@brozorec cant detect "Remove the circle shadow of checkboxes in the disclaimer modal"
Could you please tell more about it?

brozorec:
> @brozorec cant detect "Remove the circle shadow of checkboxes in the disclaimer modal" Could you please tell more about it? tried on both firefox and chrome

sure, sending you a screen recording in DM

PR #483, opened 2023-04-21
UI issues v3

NikolaiYurchenko:
closes https://github.com/Fujicracy/fuji-v2/issues/480


ferostabio:
There was a misunderstanding and we do need to do something about the "Done sir" text, plus there's a missing task about the disclaimer modal's checkmarks.

PR #486, opened 2023-04-21
Increase debounce interval

ferostabio:


brozorec:
Looks good :muscle: I compared the calls we make on the borrow page when the user is connected, from this PR and from the main:

51 (rpc) / 158 (total) vs. 51 (rpc) / 183 (total)

brozorec (approved):

PR #498, opened 2023-04-26
Vault inflation attack

DaigaroCota:
This pull request addresses vulnerabilities M-1 of Macro audit report.

### DIFF
--- a/packages/protocol/src/abstracts/BaseVault.sol
+++ b/packages/protocol/src/abstracts/BaseVault.sol
@@ -49,6 +49,7 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
   error BaseVault__mint_slippageTooHigh();
   error BaseVault__withdraw_slippageTooHigh();
   error BaseVault__redeem_slippageTooHigh();
+  error BaseVault__burn_cannotBurn();
 
   /**
    *  @dev `VERSION` of this vault.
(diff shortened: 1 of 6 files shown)


pedrovalido (commented):
Friendly reminder about the TODO's

pedrovalido:
Lets have a brainstorm session to try and attack the newly implemented fix. Action to consider:
- deposit
- withdraw
- payback
- borrow
- transfer
- burn
- malicious providers
- etc (open to more suggetions)

@DaigaroCota

pedrovalido (commented):
When a certain amount of time passes and the vault gains interest, the vault will steal the user's interest.
Flow: 
Vault has 100 shares in the beginning.
Vault has 100 fake assets in the beginning.
User deposits 100 assets
User get 100 shares
Sometime passes and the user's money gains 10 assets of interest.

User's balance would be 105 instead of 110

pedrovalido (commented):
When calling the deposit function with the vault's address as the receiver, those funds will get stuck in the router

pedrovalido:
Possible solution: when deploying a vault, use the initializeVaultShares function to also deposit a certain amount of real assets instead of faking.

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -216,6 +238,17 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+   * @dev Extended to avoid burning of address(this) shares created in
+   * `_initializeVaultShares`. See OZ implementation: {ERC20-_burn}.
+   */
+  function __burn(address account, uint256 amount) internal virtual {
+    if (account == address(this)) {
+      revert BaseVault__burn_cannotBurn();
+    }
+    _burn(account, amount);
+  }
+
```
pedrovalido, line 251: remove this. Its not going to be used

DaigaroCota:
This pull request was overriden by #541 .

PR #499, opened 2023-04-26
Rebalance breaks ltv

pedrovalido:
This PR addresses M-5

The changes made to the RebalancerManager are to prevent the rebalance to change the current ltv. So, a check was added to verify the ltv before and after the rebalance.
 Also, there's a check for redundant calls of rebalance: if the amount of assets passed = 0, then it revertes straight away. It wouldnt make sense to rebalance only debt and it would be impossible to do so and keep the same ltv

### DIFF
--- a/packages/protocol/src/RebalancerManager.sol
+++ b/packages/protocol/src/RebalancerManager.sol
@@ -144,8 +144,8 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
     {
       BorrowingVault bvault = BorrowingVault(payable(address(vault)));
       uint256 maxLtv = bvault.maxLtv();
-      uint8 assetDecimals = IERC20Metadata(bvault.asset()).decimals();
-      uint8 debtDecimals = IERC20Metadata(bvault.debtAsset()).decimals();
+      uint8 assetDecimals = vault.decimals();
+      uint8 debtDecimals = bvault.debtDecimals();
 
       //calculate ltv after rebalance at from
       uint256 assetsFrom = from.getDepositBalance(address(bvault), bvault) - assets;
--- a/packages/protocol/src/RebalancerManager.sol
+++ b/packages/protocol/src/RebalancerManager.sol
@@ -60,15 +60,17 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
       revert RebalancerManager__rebalanceVault_notValidExecutor();
     }
 
-    if (assets == 0) {
-      revert RebalancerManager__rebalanceVault_invalidAmount();
-    }
-
     _checkAssetsAmount(vault, assets, from);
 
     if (vault.debtAsset() == address(0)) {
+      if (assets == 0) {
+        revert RebalancerManager__rebalanceVault_invalidAmount();
+      }
       vault.rebalance(assets, 0, from, to, 0, setToAsActiveProvider);
     } else {
+      if (assets == 0 && debt == 0) {
+        revert RebalancerManager__rebalanceVault_invalidAmount();
+      }
       _checkDebtAmount(vault, debt, from);
       if (!chief.allowedFlasher(address(flasher))) {
         revert RebalancerManager__rebalanceVault_notValidFlasher();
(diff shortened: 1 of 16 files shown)


DaigaroCota (changes requested):

packages/protocol/src/RebalancerManager.sol
```diff
@@ -109,6 +117,22 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
...
   }
 
+  /**
+   * @dev Checks if the rebalance operation will break the current LTV
+   *
+   * @param vault address
+   * @param assets amount to rebalance
+   * @param debt amount to rebalance
+   */
+  function _checkLtvChange(IVault vault, uint256 assets, uint256 debt) internal view {
```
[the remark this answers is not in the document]
pedrovalido: Current check will be changed to check if Fuji vault will have ltv below max ltv in both provider after rebalance.

pedrovalido:
Note to self: check the assets==0 revert check. This might have to be removed due to the current path this fix is taking

packages/protocol/src/RebalancerManager.sol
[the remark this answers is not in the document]
pedrovalido: @DaigaroCota I have implemented the necessary changes. One thing I am not completely sure is the optimization of the changes I made. Can you check if its efficient?

packages/protocol/src/RebalancerManager.sol
```diff
@@ -54,6 +59,11 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
     if (!allowedExecutor[msg.sender]) {
       revert RebalancerManager__rebalanceVault_notValidExecutor();
     }
+
+    if (assets == 0) {
+      revert RebalancerManager__rebalanceVault_invalidAmount();
+    }
+
```
pedrovalido, line 66: Remove this condition. Check necessary changes to be able to rebalance just debt.

pedrovalido: Screenshot 2023-05-05 at 12 28 16
use this as an idea. ps: this is the rebalance function inside the borrowing vault.

All the _executeProviderAction should have this condition

packages/protocol/src/RebalancerManager.sol
```diff
@@ -109,6 +120,50 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
...
+    ILendingProvider from,
+    ILendingProvider to,
+    uint256 assets,
+    uint256 debt
+  )
+    internal
+    view
+  {
+    {
+      // revert RebalancerManager__checkLtvChange_invalidAmount();
```
pedrovalido, line 143: Remove this comment

packages/protocol/test/mocking/vaults/VaultRebalancing.t.sol
```diff
@@ -442,4 +466,27 @@ contract VaultRebalancingUnitTests is MockingSetup, MockRoutines {
...
+
+    bool rebalanceLtv =
+      _utils_checkRebalanceLtv(bvault, mockProviderA, mockProviderB, rebalanceAssets, rebalanceDebt);
+
+    //rebalance
+    try rebalancer.rebalanceVault(
+      bvault, rebalanceAssets, rebalanceDebt, mockProviderA, mockProviderB, flasher, true
+    ) {
+      if (rebalanceLtv) {
+        assert(true);
```
pedrovalido, line 484: change assert to assert(rebalanceLtv)

packages/protocol/src/RebalancerManager.sol
```diff
@@ -109,6 +120,50 @@ contract RebalancerManager is IRebalancerManager, SystemAccessControl {
...
+      BorrowingVault bvault = BorrowingVault(payable(address(vault)));
+      uint256 maxLtv = bvault.maxLtv();
+      uint8 assetDecimals = IERC20Metadata(bvault.asset()).decimals();
+      uint8 debtDecimals = IERC20Metadata(bvault.debtAsset()).decimals();
```
pedrovalido, line 147: Read decimals from bvault

pedrovalido:
All above changes have been addressed. Let me know if it ready to merge @DaigaroCota

PR #508, opened 2023-05-05
M-4 Max function compliance with EIP4626

pedrovalido:
This PR addresses M-4


packages/protocol/src/vaults/borrowing/BorrowingVault.sol
DaigaroCota, line 1: Hey @pedrovalido, in recent merge we introduced more debt-equivalent functions that should also be considered to return zero when paused. This include: `maxPayback`, `maxMintDebt`, and `maxBurnDebt`.
Please include the changes to these too.

DaigaroCota (changes requested):

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
pedrovalido: changes applied :)

PR #513, opened 2023-05-09
From token to currency

### LINKED ISSUE #173: Use Currency instead of Token
- blocks #174 

We may need to refacto `Token -> Currency` in order to support Native assets.
Need to be investigated.

### LINKED ISSUE #174: Borrow: skip allowance for native assets
- blocked by #173

### Context

We are about to support native assets (i.e eth for Ethereum, Matic for polygon...).
If the user wants to borrw with a native assets, then we have to skip the allowance steps.

### Todo

- [ ] If collateral `Token.isNative`, skip allowance button

### Explanations

To check if a token is native, use `Token.isNative` (ez).

ferostabio:
close #173 
close #174


ferostabio:
@NikolaiYurchenko we've been doing some cleanup, in order to continue with that, please create a branch with this as base and move all the `acceptTermsOfUse`, `getOnboardStatus`, `setExploreInfoSkipped` and related functions to the `auth` helper since they don't make any use of the `store`.

brozorec:
Some feedback:

- When I land directly on the manage position, the active vault is always the one on Polygon even though the URL is the address of another one
- I want to mange my WETH/USDC position on Arbitrum and I switch WETH -> ETH, the active vault got changed.
- Similar to the above: I'm on the borrow page and Arbitrum as selected chain. I select from the routes the second vault, then I switch WETH -> ETH and the active vault is changed back to the first one.

While debugging, I came across the logic of how we set when the user has a balance in the active vault. With the new returned type `VaultsWithFinancails`, we have this data so there's no need to fetch it separately. @ferostabio can you have a look, do I miss anything there?

PR #520, opened 2023-05-11
Positions table row providers info

NikolaiYurchenko:


ferostabio:
@NikolaiYurchenko you're showing only one provider instead of all. Tooltip says `In the background, Fuji rebalances between these protocols to provide the best terms.` but there's only one, the active one, when we can move things around. Needs to be like the Markets screen.

ferostabio (approved):

PR #521, opened 2023-05-11
Login improvements

ferostabio:
There were two issues in the login flow:

- `Init` / `reconnect` were called twice.
- There was code in `onOnboardChange` that triggered a logout every time the app loaded, so the `Connect wallet` button flashed for a bit.

Tested a number of times, but we should all do it a number of times in case there are any side-effects or something I'm missing.

### DIFF
--- a/packages/frontend/store/auth.store.ts
+++ b/packages/frontend/store/auth.store.ts
@@ -1,9 +1,5 @@
 import { ConnectOptions } from '@web3-onboard/core';
-import {
-  Balances,
-  ConnectedChain,
-  WalletState,
-} from '@web3-onboard/core/dist/types';
+import { Balances, WalletState } from '@web3-onboard/core/dist/types';
 import { ChainId } from '@x-fuji/sdk';
 import { ethers, utils } from 'ethers';
 import produce from 'immer';
@@ -11,7 +7,7 @@ import { create } from 'zustand';
 import { devtools } from 'zustand/middleware';
 
 import { web3onboard } from '../helpers/auth';
-import { chainIdToHex } from '../helpers/chains';
+import { chainIdToHex, hexToChainId } from '../helpers/chains';
 
 export const onboard = web3onboard;
 
@@ -33,7 +29,7 @@ type AuthState = {
   address?: string;
   ens?: string;
   balance?: Balances;
-  chain?: ConnectedChain;
+  chainId?: ChainId;
   provider?: ethers.providers.Web3Provider;
   walletName?: string;
 };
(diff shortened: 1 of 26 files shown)


ferostabio:
After doing the first batch of (minor, granted) changes, address still took an average of 1.75 seconds to load, when all major projects load it almost instantly.

I went through `@web3-onboard`'s rabbit hole and couldn't find any way to speed up `onboard.connectWallect()`. I started looking at the different sub packages of them we already use, and realized that they offer a `useWallets` hook that, when implemented inside a component, it shows the connected wallets right after loading.

A couple of comments:

- We weren't saving the ens name during the connection, only afterwards.
- The hook can't be called before initiating `@web3-onboard`, reason I added the property and safety checks.
- I replaced the `onOnboardChange` with the hook.
- I took the chance to clean the `auth.store`a bit.

packages/frontend/store/auth.store.ts
```diff
@@ -20,37 +21,21 @@ type OnboardStatus = {
   isExploreInfoSkipped?: boolean;
 };
 
-type ConnectedState = {
-  status: 'connected';
-  address: string;
+type AuthState = {
+  status: 'initial' | 'connected' | 'disconnected';
```
brozorec, line 25: We better use an enum instead of strings. That will require refactoring in other places but as we're dealing with this now, we better push it one step further.

packages/frontend/store/auth.store.ts
```diff
@@ -20,37 +21,21 @@ type OnboardStatus = {
...
+  address: string | undefined;
   ens: string | undefined;
-  balance: Balances;
-  chain: ConnectedChain;
-  provider: ethers.providers.Web3Provider;
-  walletName: string;
+  balance: Balances | undefined;
+  chain: ConnectedChain | undefined;
+  provider: ethers.providers.Web3Provider | undefined;
+  walletName: string | undefined;
```
brozorec, line 31: what's the advantage of using `address: string | undefined` vs `address?: string`?

packages/frontend/store/auth.store.ts
```diff
@@ -61,16 +46,17 @@ type Action = {
...
 
-const initialState: InitialState = {
+const initialState: AuthState = {
   status: 'initial',
   address: undefined,
   ens: undefined,
   balance: undefined,
   chain: undefined,
   provider: undefined,
   walletName: undefined,
```
brozorec, line 58: no need to explicitly assign `undefined` values

brozorec (changes requested):

packages/frontend/store/auth.store.ts
ferostabio: None I can think of, quick refactor leftover

packages/frontend/store/auth.store.ts
ferostabio: Yes, another quick refactor leftover

packages/frontend/store/auth.store.ts
ferostabio: Agree!

PR #525, opened 2023-05-12
Markets store

### LINKED ISSUE #517: Market store
Create a `market.store` and keep market data there to prevent it from being lost every time the user enters the markets page.

- [ ] Create `market.store` and move functions there
- [ ] Update `README.md` file

ferostabio:
close #517

Created a `markets.store`,  moved there all functions from `MarketsTable`. In order to avoid making a huge refactor I also moved the row creation code there, so the store has a `rows` convenience property to be used in the markets page (a lot of work would be needed otherwise in order to break things down more, including all of the components).

Anyway, took the chance to make some minor improvements to the helper and stores as well.

### DIFF
--- a/packages/frontend/README.md
+++ b/packages/frontend/README.md
@@ -127,6 +127,7 @@ stores
 ├── auth
 ├── borrow
 ├── history
+├── markets
 └── positions
 ```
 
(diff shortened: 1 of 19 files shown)


ferostabio:
@brozorec there is one more thing to do here.

At the moment, when the user returns to the page, we re fetch the data (of course), and that means we re set the initial data and we show the loaders for the financials. Give it a try and see.

In order to avoid this, we need to match the initial call's data and only update the UI (rows) if something basic changed (we're still loading the financials again, and we will set the updated info, only we want to avoid showing the loaders).

Danger is, if we're shallow in the comparison and something did change and the user taps the row during that time, he'll have outdated data. So we need to be sure about what we check (not everything, but not just what we show).

PR #526, opened 2023-05-15
Refactor borrow store

brozorec:
- deprecate `LendingProviderDetails` in favor of the `LendingProviderWithFinancials`
- refactor `sdk.getBorrowingVaultsFor` to use `batchLoad` in order to limit the number of requests
- stop fetching separately providers, as they are already available
- remove component `VaultsMenu` as we don't use it anymore as it was intended initially
- remove `updateAllProviders` from the store because we don't need all providers for the available vaults anymore


brozorec:
Ideas for further improvements:

- adopt a better naming: it's not clear what's the difference between "updateSmth" and "changeSmth" methods
- formatting to improve readability: create a separate const `store` to use here instead of passing the whole store object
```
export const useBorrow = create<BorrowStore>()(
  persist(devtools(store, { ... }))
);
```
Use typings similar to "auth.store"

ferostabio (approved):

PR #536, opened 2023-05-17
Macro post review critical-high

DaigaroCota:
This pull request addresses the issues identified during the post review by Macro of critical and high findings.


packages/protocol/src/routers/ConnextHandler.sol
```diff
@@ -131,16 +150,16 @@ contract ConnextHandler {
     external
     onlyConnextRouter
   {
-    if (!isTransferIdRecorded(transferId)) {
-      _failedTxns[transferId] =
-        FailedTxn(transferId, amount, asset, originSender, originDomain, actions, args);
-    }
+    uint256 nextNonce = getFailedTxnNextNonce(transferId);
+    _failedTxns[transferId][nextNonce] =
+      FailedTxn(transferId, amount, asset, originSender, originDomain, actions, args, nextNonce);
```
brozorec, line 155: @DaigaroCota can we add here an event with the same params? This will greatly facilitate the tracking of failed txns as we are only going to deal with this contract without the need to refer to "ConnextRouter".

packages/protocol/src/routers/ConnextHandler.sol
```diff
@@ -152,18 +171,18 @@ contract ConnextHandler {
...
     external
     onlyAllowedCaller
   {
-    FailedTxn memory txn = _failedTxns[transferId];
+    FailedTxn memory txn = _failedTxns[transferId][nonce];
     IERC20(txn.asset).approve(address(connextRouter), txn.amount);
-    try connextRouter.xBundle(txn.actions, args) {
-      delete _failedTxns[transferId];
-      emit FailedTxnExecuted(transferId, true, txn.actions, actions, txn.args, args);
+    try connextRouter.xBundle(actions, args) {
```
brozorec, line 182: Once the tx gets successfully executed, should we mark it somehow in order to prevent replaying it?

packages/protocol/src/routers/ConnextHandler.sol
DaigaroCota: Requested event was added.

packages/protocol/src/routers/ConnextHandler.sol
DaigaroCota: the struct for failed txs now includes an boolean argument for "executed?". Calling an executed failed transferId will revert.

PR #537, opened 2023-05-17
Borrowing flow

### LINKED ISSUE #573: New borrowing flow

ferostabio:
close #573 

- [x] hide "estimated cost" in initial state
- [x] don't show 0 values in routes (check current implementation)
- [x] tapping in routes selects the vault
- [x] keep only one expanded vault (tapping in "see routes" for other vault collapses the currently expanded one)
- [x] animate initial state + overview transition 🎉
- [x] animate unfolding 
- [x] hover on both rows
- [x] folded container
- [x] fees
- [x] place box on right
- [x] rework widget (make containers)
- [x] mobile adaptations


ferostabio:
It’s a very, very early draft, a starting point, but wanted to push as soon as possible so, with a couple of tweaks if needed, @NikolaiYurchenko can already start working

- As I mentioned above it's a draft. `borow.store`'s `debt` property is now `undefined` by default and quickly updated all usages. UI mostly just checks for undefined and returns either `null` or `<>`.
- Most times, both UI and store, checking if we there's a `debt` value is more than enough.
- There are a couple of places in which it might not be enough, at all. Added lots ot `TODO` in places we need to think about
- The `Position` object is a good example. In the `positions` helper's `viewDynamic…` function, it expects a `debt` object. And the user won't be managing a position and have no debt, so there's no point in allowing there an optional field. Nothing big, but need to figure it out.

ferostabio:
- [ ] hide "estimated cost" in initial state
- [ ] don't show 0 values in routes (check current implementation)
- [ ] tapping in routes selects the vault
- [ ] keep only one expanded vault (tapping in "see routes" for other vault collapses the currently expanded one)
- [ ] animate initial state + overview transition 🎉

ferostabio:
- [ ] Show current position in the borrowing screen, not just when managing a position
- [ ] Jumping beta bar seems to be back

https://github.com/Fujicracy/fuji-v2/assets/501915/2633548a-29ef-4772-9c15-bd2d627b97fb

ferostabio:
- [ ] weth -> eth before choosing a chain doesn't work (Federico)
- [ ] increase animation time
- [ ] undefined.content (notification)
- [ ] going back to borrow after choosing a debt, the screen glitches (shows overview before disappearing, need to make sure to remove the debt before we load the data)
- [ ] add overview title
- [ ] remove blank space between amount and percentage in routes
- [ ] route hover, left state has rounded corners but not on the right
- [ ] details text in overview's bottom has not enough top margin
- [ ] fold routes when selecting one
- [ ] typing an input (and tapping max/recommended) folds routes
- [ ] also happens changing chain/token
- [ ] change wording from 'chain' to 'network'
- [ ] change "collateral apy" to "supply apy" in markets (Federico will change it in the apy chart)
mobile:
- [ ] all vaults is too close to the left
- [ ] icons are above the overview button

brozorec:
- [ ] When switching tokens or networks, the list of "All Vaults" can get updated. Add a skeleton loader so that the change is more evident.
- [ ] "See all vaults" is with a smaller font size, rename "Close" to "Hide".
- [ ] Add animation when folding and unfolding.
- [ ] UI issue 
image

- [ ] When I directly hit "/my-positions/VAULT_ADDR-CHAIN_ID", the page doesn't load.
- [ ] When I'm on manage position, I can't do a deposit-only or a withdraw-only.
- [ ] I want to make a payback+withdraw and there's an approval. After the approval tx gets executed, no signature gets prompted. There's an error notification and the button doesn't stop loading.
- [ ] I want to withdraw only. Approval shouldn't be needed:
image
After I approve (although I shouldn't be asked for), the button displays "Signs & Borrow", I guess there's an issue with the `mode`:
image

brozorec:
- [ ] I still have the issue "I want to make a payback+withdraw and there's an approval. After the approval tx gets executed, no signature gets prompted." and the tx fails.
- [ ] In manage position, the debt (both when borrowing and when payback) is not taken into account in the LTV gauge and when hitting the max withdraw button.
- [ ] I'm on the borrow page and the current chain is different than the collateral chain. When I change the chain, the form re-init.
- [ ] Change "Select a chain" to "Select a network" and "Select a chain to borrow on" to "Select a network to borrow on"

ferostabio:
>  I'm on the borrow page and the current chain is different than the collateral chain. When I change the chain, the form re-init.

I either cannot reproduce or don't fully understand this one @brozorec

ferostabio:
- [ ] Align borrow form
- [ ] Vault selection’s available route mismatch, crash
- [ ] Transitioning Borrow -> Position clears the overview panel, poor transition

PR #542, opened 2023-05-18
Lending markup (old)

### LINKED ISSUE #545: Lending: markup and preparations

NikolaiYurchenko:


ferostabio:
Some important points to tackle once that the branch is up to date.

- The SDK changed, there’s no `LendingProviderDetails` anymore so the `rearrangeWith…` function needs to be refactored (ps: I moved the function to a new `lending` helper, we shouldn’t add lending functions to any borrowing related code).
- `viewDynamicPosition` and `showPosition` changed, params are different and can now be undefined, so the flow needs to be updated and double checked.
- Reintroduce the commented out components, all of them were changed this month, so please update them.
- Remove duplicated code, and create reusable components whenever possible. Rows and tables are currently sharing a lot of functionality and most code has just been duplicated (`loaderOrError` is a small but good example).
- Get rid of all `any`, please use types whenever possible
- Rename components like `BorrowBox` if they’re going to be used in Lending. This type of component won’t use `borrow.store` but a new `lending.store`. It’s not just a matter of doing the markup, we have to be thorough in the component structure. It makes sense to create a dumber `FormBox` component, for instance, that accepts data that will be sent from `BorrowBox` and `LendingBox`.

brozorec:
moved all related frontend work to #631

PR #550, opened 2023-05-22
Routes redesign

NikolaiYurchenko:
Routes stepper styles rework
Assets Box rework to reuse in Routes modal
Confirmation transaction modal redesign
Tx modal look similar


ferostabio:
@NikolaiYurchenko I would say continue in this PR and update the tx history. You can use the same modal, only for the time being at least maintain the step grouping

brozorec:
1. I manage a position on Optimism and I payback from Gnosis. The "from" titile should be Gnosis and not Optimism.
image

2. I'm not sure about the sign "->" below "Deposit to Aave V3 and Borrow from Aave V3". It's for the bridging, isn't "0.001 WETH / 1.2 USDC" better?
image

3. Based on the above, when we have double steps "deposit+borrow" or "payback+withdraw", there's a redundant provider name. So the wording be better "Deposit and borrow from Aave V3" and "Payback and withdraw from Aave V3".

NikolaiYurchenko:
Thanks, will do. Only don't sure about second, for me it is pretty clear, Ill change and show how it will be

ferostabio:
> Thanks, will do. Only don't sure about second, for me it is pretty clear, Ill change and show how it will be

Agree with @brozorec on this one. Find the "->" confusing. I like how it looks, but I think it gives the impression of a swap: WETH is exchanged for USDC, and that's not what's happening.

ferostabio:
Is there a reason for having this when there's a bridge? We're repeating the amount, we're literally doing `${singleAmount} -> ${singleAmount}`. I would just show the amount we're bridging @brozorec 

Captura de pantalla 2023-05-24 a la(s) 14 32 27

ferostabio (approved):

PR #569, opened 2023-05-30
Onboarding changes

NikolaiYurchenko:
Disclaimer only after connect wallet click
changed behavior, tests


ferostabio (changes requested):
Great work @NikolaiYurchenko !

One small request: change the code so all the components call auth.store's `login` function, just like it was initially, and the `login` function checks wether we've shown the permission or not. That way it will be more clear. Specially considering the user can accept the disclaimer and then tap 'Connect wallet' again, so calling the `showDisclaimer` function is a bit weird.

PR #578, opened 2023-06-05
Guild.xyz to determine eligibility

brozorec:


packages/frontend/components/Access/GuildAccess.tsx
```diff
@@ -0,0 +1,188 @@
...
+      }, 3000);
+    } else {
+      setIsOpen(true);
+    }
+  }, [accessStatus]);
+
+  const statusParagraph = () => {
+    if (authStatus !== AuthStatus.Connected) return <></>;
+
+    if (accessStatus === AccessStatus.Verifying) {
```
NikolaiYurchenko, line 59: better to use content table for this aka
const statusLabels = { AccessStatus.Verifying: 'Verifying your eligibility ...',  }
and get string like this statusLabels[accessStatus]

packages/frontend/components/Access/GuildAccess.tsx
```diff
@@ -0,0 +1,188 @@
...
+              size="medium"
+              fullWidth
+              loading={authStatus === AuthStatus.Connecting}
+              onClick={() => login()}
+            >
+              Connect Wallet
+            </LoadingButton>
+          ) : (
+            <Stack direction="row" justifyContent="space-between" spacing={3}>
+              <Tooltip title={statusParagraph()} placement="top" arrow>
```
NikolaiYurchenko, line 126: better to use const wrapped with useMemo

packages/frontend/components/Access/GuildAccess.tsx
```diff
@@ -0,0 +1,188 @@
...
+  return (
+    <Dialog
+      open={isOpen}
+      /* BackdropProps={{ style: { backgroundColor: "" } }} */
+      sx={{
+        '.MuiPaper-root': {
+          width: isMobile ? '100%' : 'auto',
+        },
+        '&': {
+          // in order for onboard modal to display above this dialog
```
NikolaiYurchenko, line 82: maybe it is better to show them in order?

ferostabio (approved):

PR #606, opened 2023-06-19
Handle 0 fees and slippage on cross tx

NikolaiYurchenko:
Did not changed sdk behaviour because catching and formatting data on FE seems like bad idea. We already have needed conditions (crosschain and 0 data)


ferostabio:
> Did not changed sdk behaviour because catching and formatting data on FE seems like bad idea. We already have needed conditions (crosschain and 0 data)

At the moment, the SDK is returning `success` instead of `error` when the call to `Connext` fails. Can't think of anything worst than that! What happens is that for this particular use case, we want to keep going even if the call fails. So while I understand your concern, I don't think it's a problem at all.

Thanks for doing the work, will continue with this my self!

brozorec (approved):

PR #612, opened 2023-06-19
Handle chain change error

### LINKED ISSUE #607: Change chain error
We're getting an error in production we never saw during development

```
Error: A wallet must be connected before a chain can be set
  at <object>.changeChain (./store/auth.store.ts:130:23)
  at selectChain (./components/App/Header/ChainSelect.tsx:40:5)
  at <object>.onClick (./components/App/Header/ChainSelect.tsx:94:28)
...
(11 additional frame(s) were not displayed)
```

ferostabio:
close #607

### DIFF
--- a/packages/frontend/store/auth.store.ts
+++ b/packages/frontend/store/auth.store.ts
@@ -7,6 +7,8 @@ import { devtools } from 'zustand/middleware';
 
 import { getOnboardStatus, web3onboard } from '../helpers/auth';
 import { chainIdToHex, hexToChainId } from '../helpers/chains';
+import { stringifyError } from '../helpers/errors';
+import { notify } from '../helpers/notifications';
 import { storeOptions } from '../helpers/stores';
 
 export const onboard = web3onboard;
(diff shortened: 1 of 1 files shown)


ferostabio:
Error happens in `web3-onboard/core`’s library, in the `chain.ts` file. When the `setChain` function is called, if there’s no wallet, it throws an error:

```ts
if (!wallet) {
    throw new Error(
      walletToSet
        ? `Wallet with label ${walletToSet} is not connected`
        : 'A wallet must be connected before a chain can be set'
    )
  }
```

The only mention I could fine was [this](https://github.com/blocknative/web3-onboard/issues/938), although I’m afraid it’s of no use to us.

We never saw this issue while developing and I’ve been completely unable to reproduce, but it already happened a couple of times.

For the time being, I’m wrapping our `await onboard.setChain({ chainId: hexChainId })` function call in a try/catch block, to show the user a notification with the error instead of simply failing.

brozorec (approved):

PR #621, opened 2023-06-22
Reentrancy in connext handler failed tx execution

DaigaroCota:
This pull request addresses Item #-23 of Three Sigma security audit.


0x73696d616f (approved):
Looks good

packages/protocol/src/routers/ConnextHandler.sol
```diff
@@ -195,13 +195,16 @@ contract ConnextHandler {
       revert ConnextHandler__executeFailed_tranferAlreadyExecuted(transferId, nonce);
     }
 
+
     SafeERC20.safeIncreaseAllowance(IERC20(txn.asset), address(connextRouter), txn.amount);
+    txn.executed = true;
+    _failedTxns[transferId][nonce] = txn;
```
pedrovalido, line 201: please address the changes in PR #616

packages/protocol/src/routers/ConnextHandler.sol
```diff
@@ -195,13 +195,16 @@ contract ConnextHandler {
...
 
     try connextRouter.xBundle(actions, args) {
-      txn.executed = true;
-      _failedTxns[transferId][nonce] = txn;
       emit FailedTxnExecuted(transferId, txn.actions, actions, txn.args, args, nonce, true);
     } catch {
+      txn.executed = false;
+      _failedTxns[transferId][nonce] = txn;
```
pedrovalido, line 207: same here

pedrovalido (changes requested):

packages/protocol/src/routers/ConnextHandler.sol
DaigaroCota: Hey Pedro, per our offline discussions I believe I was wrong on my point. 
In `_failedTxns[transferId][nonce]`, the `nonce` is used to access the specific position in the array of struct array  FailedTxn[] of the _failedTxns mapping.
`mapping(bytes32 => FailedTxn[]) private _failedTxns;`
Therefore it is okay as is.

DaigaroCota: I only added clarifying notes for the `nonce` argument in the functions that have it. So that is not mis-interpreted.

packages/protocol/src/routers/ConnextHandler.sol
DaigaroCota: See comment above.

packages/protocol/src/routers/ConnextHandler.sol
pedrovalido: Makes sense :)

pedrovalido (approved):

PR #623, opened 2023-06-22
Replace safe approves

DaigaroCota:
This pull request addresses Item #-6 of the Three Sigma security audit.


packages/protocol/src/routers/ConnextHandler.sol
```diff
@@ -216,7 +216,7 @@ contract ConnextHandler {
       revert ConnextHandler__executeFailed_tranferAlreadyExecuted(transferId, nonce);
     }
 
-    IERC20(txn.asset).approve(address(connextRouter), txn.amount);
+    SafeERC20.safeIncreaseAllowance(IERC20(txn.asset), address(connextRouter), txn.amount);
 
     try connextRouter.xBundle(actions, args) {
```
0x73696d616f, line 221: Think the allowance should be decreased if the `xBundle(...)` call fails (assets would not have been pulled)

0x73696d616f (changes requested):
Just left 1 note, everything else looks good

packages/protocol/src/routers/ConnextHandler.sol
DaigaroCota: Added decrease allowance in catch as suggested, considering failed execution.

0x73696d616f (approved):
Looks good

pedrovalido (approved):

PR #631, opened 2023-06-26
Lending Vaults

brozorec:
#### Main tasks

- [x] General markup
- [x] Lending vaults in the sdk plus related functions
- [x] Lending store
- [x] Vaults helper
- [x] Lending markets
- [x] lend market returns 0 items
- [x] on Lend page I got infinite FujiResultErrors
- [x] Lending history
- [x] Lending positions
- [x] Abstracting operation logic (like activeVault, since now we’ll have two)
- [x] Update my-positions/[pid] page so the url contains either “borrow” or “lending”
- [x] Update navigation accordingly
- [x] Update positition.store calculations taking into account borrow/lending
- [x] Reuse position components like YieldTable since they’re mostly the same, they need to accept a positions parameter so we can pass either borrowPositions or lendingPositions -that’s just an example, we can’t be thorough with this but can’t be sloppy either
- [x] Reuse position components MyPositionTable
- [x] UI: Show lending markets and redirect to corresponding screens (using showLend, showLendingPosition functions)
- [x] UI: Show lending positions (and direct to…)
- [x] UI: Show different components in my-positions/[pid] screen depending on the url
- [x] UI: Connect lending screen with the lending store
- [x] UI: Manage lending position (connect to store)
- [x] UI: Transaction modal for lending operation // Started
- [x] UI: Show lending operations in history
- [x] UI: Allow showing position list defaulting to lending 

#### Issues

- [x] Use VaultType as VaultSelect param.
- [x] If the user taps lending markets before they’re loaded, borrowing markets are shown
- [x] Create a common row for market tables. Not everything needs to be shared, but while updating types yesterday had to do some things twice (Nick)
- [x] Update market query params so instead of tab 0/1 its borrow/lending, for consistency's sake
- [x] Fix warning: `Image with src "/assets/images/protocol-icons/networks/Optimism.svg" has legacy prop "objectFit". Did you forget to run the codemod? Read more: https://nextjs.org/docs/messages/next-image-upgrade-to-13`
- [x] At the moment there's a problem fetching the vaults, anyway the lending button screen is disabled but with `Unsupported pair` text.
- [x] At the same time, chart isn't show at all. Sharing an image to describe both points.

Captura de pantalla 2023-07-06 a la(s) 11 14 30

- [x] Filter vault selection component by type 
- [x] Wrong vault passed / returned around from markets to lending (Federico)
- [x] Hide Borrow Apr in VaultSelect for lend case
- [x] my deposit and remove rewards at the top right
- [x] move current apy and total supply on top of chart
- [x] replace current apy with my deposits
- [x] show total supply in eth and wait for sdk response to add all lendingProviders values
- [x] Collateral defaults to Polygon, so borrowing mechanism needs to be copied (Federico) 

**Known issues**
- [x] after first deposit nothing happens (should redirect to positions maybe?)
- [x] poor amount rounding (deposited 0.001 eth, withdrawed 0.0005 and position says that I have 0.001 deposited)
- [x] after full withdraw no updates on page (should update positions)
- [x] revert Balance in Assetbox 
- [x] mobile vault select changes for lend
- [x] initial chart doesn't show data

https://github.com/Fujicracy/fuji-v2/assets/501915/99815b4c-7cd0-4c97-8875-b36b8054a6d4

- [x] Change "Lend APY" to "Supply..."
- [x] Don't change vaults when changing chains in lending (Federico)
- [x] Update available vaults in the sdk so it returns all vaults in all chains for a given token (Federico)


ferostabio:
- There's some code duplication with the new `lending.store` but not logic. Tried to improve this but a higher-function that returns `SharedAction` type just like we pass the initial state is by far worst: types are a nightmare and we end up using `any`, loosing clarity. 
- Considered adding `changeChain` style functions in order to streamline the `lend.store` interface, but it would end up adding complexity to our components, since they mostly expect a general interface with an `AssetType` param and it would end up creating more trouble. In consequence, I added safety checks to make sure nothing bad happens. Feel free to suggest improvements.
- Added `TODO` messages to remember key code that needs to be updated, going to go back to the `Sdk` tomorrow and will go back once we're done

ferostabio:
I added a checklist item to the PR, focusing on what remains to be done. I’m sure some of the items marked as done might need some work and/or tweaks.

@NikolaiYurchenko , please start working on them top to bottom -and please let me know if you spot any issues or missing code in any of the check list items.

packages/frontend/components/Borrow/VaultSelect/VaultSelect.tsx
```diff
@@ -19,10 +19,11 @@ import { useTheme } from '@mui/material/styles';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { useBorrow } from '../../../store/borrow.store';
+import { useLend } from '../../../store/lend.store';
 import { useNavigation } from '../../../store/navigation.store';
 import Vault from './Vault';
 
-function VaultSelect() {
+function VaultSelect({ type = 'borrow' }: { type?: 'borrow' | 'lend' }) {
```
brozorec, line 26: we better use here the type `VaultType`

packages/frontend/store/models/Position.ts
brozorec, line 1: I was also thinking along those lines about the types :+1:

ferostabio:
Want to leave the description for the main issues, so I'm adding a new list with issues @NikolaiYurchenko 

- [ ] Use VaultType as VaultSelect param.
- [ ] Providers not shown for lending markets
- [ ] If the user taps lending markets before they’re loaded, borrowing markets are shown
- [ ] Create a common row for market tables. Not everything needs to be shared, but while updating types yesterday had to do some things twice

ferostabio:
Just removed an item from the checklist. While in order to fully develop both positions and managing position screens we do need to be able to make at least a deposit, a lot of work can be done before then having the borrowing counterpart as example.

ferostabio:
Added a couple of issues to the list @NikolaiYurchenko !

@brozorec we're having a problem in which...
1 - We get lending markets with `sdk.getAllLendingVaults()`.
2 - User taps one. Tiggers `changeAll`, which calls `await sdk.getLendingVaultsFor(collateral, account)`.

Vault returned in point 1 is not found on point 2.

Is this caused by not having the right vaults deployed or is there a problem with some of the functions? I haven't found a bug so far, but 2) uses batchLoad. We should deploy vaults as soon as possible.

ferostabio:
Hey @NikolaiYurchenko ! In the last two commits (33e3c7569229362bdc0b1b20d280b3910599a672 and 7e90c4e78ab81844acbe82ef3ebe4cd24ce08489) you removed the `position` object from Lending related components.

You changed `LendingButton`’s line 116 from `collateralAmount > Number(position?.collateral?.amount)` to `collateralAmount > Number(collateral?.amount)` first and then completely removed the `position` property.

While there is no need for the position object while doing the lending screen, it will be once we manage the position and lending should replicate the borrowing counterpart as much as possible. Idea is to make it simple to detect issues in any of the parts and knowing what’s going on in one of them after looking at the other one.

So please revert all this changes before its late and make sure lending works like borrowing (passing the position around).

ferostabio:
@brozorec had to update the `batchLoad` function since it expected a `chain` argument, meaning we were only able to get `availableVaults` for a single chain, please have a look at the new `multiBatchLoad` function.

Also, the reason we're not showing a provider icon for Optimism's lending vault is becaue we just don't have the data, this is what we have the `sdk`in the `allProviders` property, in `batchLoad`. No `name` and no `llamaKey`, just `depositAprBase` for one of the elements.

```js
[
  {
    "depositAprBase": 0.5913637758368
  },
  {
    "name": "Aave V3",
    "llamaKey": "aave-v3",
    "depositAprBase": 1.1658847850287968
  }
]
```

ferostabio:
@NikolaiYurchenko please have a look! 

When loading `/lend` directly:

- Vaults component don't show loading state, so it seems like there's an issue
- Until loading we just `My deposits`, no `Total supplied`
- After loading, both components show `(undefined)`

https://github.com/Fujicracy/fuji-v2/assets/501915/37f954ac-8f4c-40f0-9cd8-5f1a01669414

ferostabio:
In `/my-positions`, when going from borrow to lending positions, we show the loading skeletons and take a while to show the list. We actually fetch both borrowing and lending positions at the same time, so we should be able to show lending positions instantly

ferostabio:
Hey @NikolaiYurchenko some more lending tweaks. Not adding them anymore to the main list since it was getting out of control.

- [ ] Implement in Lending Wrapper same loading mechanism as in Borrowing Wrapper
- [ ] Make (!) a bit smaller when there's no total supplied amount
- [ ] Mix InfoBlock with the component we use in Borrowing that allows to see the edited amount ("after..."). Add the component if you want, I'll make sure to show the proper data

ferostabio:
@brozorec 

we are getting this A LOT while working on the lending branch

Captura de pantalla 2023-07-17 a la(s) 16 04 56

ferostabio:
We have no way of knowing the current vault's chain when managing a lending position. We're not showing the icon as we do when borrowing. In this screenshot I'm attaching, I want to deposit from Optimism to a Polygon position. When creating the position we have the vault selection component at the top left, but now we don't (when managing a borrowing position, we show the icon above the input fields).

What should we do @Markoyw ?

Captura de pantalla 2023-07-18 a la(s) 11 03 12

ferostabio:
> Hey @ferostabio
> 
> Good point, I didn't take that into account.
> 
> At the moment, I can't think of anything better than to keep it consistent with how we are displaying on the borrowing panel.
> Screenshot 2023-07-18 at 10 11 16 PM

I agree, @NikolaiYurchenko pls implement it!

ferostabio:
@NikolaiYurchenko above made me realize we weren't even showing the slippage component, when the user could be doing cross-chain lending both in creation and managing screens

ferostabio:
@brozorec about the 

```reverted transaction error while fetching `Error: missing revert data in call exception; Transaction reverted without a reason string... error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32017,\"message\":\"Timeout\",\"data\":\"Nethermind.JsonRpc.Modules.ModuleRentalTimeoutException: Unable to rent an instance of IEthRpcModule. Too many concurrent requests.```

error we're getting a lot, no idea why we're getting it so many times, but it seems to be a known [issue](https://github.com/NethermindEth/nethermind/issues/5923)

brozorec:
Here are some issues:

- trailing 0 when there is no position yet
image

- double display when managing
image

- change wording to "..., please select a different vault."
image

ferostabio:
> Here are some issues:

On it!

brozorec:
Good job in pimping the UI according to the proposed design system. Here are several more issues/points:

- [ ] On manage position, in **withdraw** mode, the "Switch to CHAIN" doesn't correspond to the selected chain in "Withdraw to" dropdown
image
- [ ] On manage position, 1) the "after" label is with a neg sign when withdrawing max; 2) I suggest we remove the tooltip from "My deposits", having only the title is explicit enough
image
- [ ] When I do a x-chain deposit, 1) the "Est. cost" doesn't get populated; 2) nothing happens when clicking the deposit btn
image
- [ ] Displaying issue in the modal for lending
image
- [ ] Comment out the "Risk" section, we are not ready with the risk framework yet
image
- [ ] Replace "Current Active Provider:" by "Underlying money markets" (without the colon ":" )
image
- [ ] Fix the fonts in the "Vault Strategy" section and replace the text with the following:
"Automatic interest rates optimization
The vault aims to maximize the returns on its assets by seeking the highest yields available from the list of aggregated lending protocols. This strategy is best for users who are "Lenders-only". Whenever they need to borrow against their provided assets, they can switch from a lending to a borrowing vault (the migration tool will be available soon).
💡️ It's important to note that each vault aggregates its own set of lending/borrowing protocols. The protocols illustrated above may vary from one vault to another."
- [ ] Missing header in manage position
image
- [ ] Change column titles in "My positions" to: "Supplied Amount" and "Supply APY"
image

brozorec:
@ferostabio @NikolaiYurchenko I fixed the fetching issues for gnosis (was a dumb typo)

- [ ] (from the previous batch of comments) I suggest we remove the tooltip from "My deposits", having only the title is explicit enough
- [ ] I suggest setting the default period on all charts to be 30d instead of 7d (the 30d chart has more data, and the curves look better)
- [ ] Adopt formatting similar to the "Liquidity" column in "Markets"
image

ferostabio:
@ansteadm discovered a bug: we are only showing the right `supply apy` in the markets section. We're showing wrong values both in "All vaults" (inside the lending screen) and my positions list.

I already fixed a couple of issues on the positions screen and now it shows the right data, but we still have a couple of issues here on the lending screen

- First, @NikolaiYurchenko , we're simply showing a string value, we need to use the same `APRValue` component we use in the markets tableview row.

- Second, and more important @brozorec , we don't have the data. 

In the lending screen we use availableVaults to show data and here `console.log(availableVaults.map((v) => v.activeProvider?.depositAprReward));` returns `[0,0, undefined, null]` -when two values do have rewards.

PR #636, opened 2023-06-27
Borrowing vault liquidate with close factor arg

DaigaroCota:
This pull request addresses Item #-47 of the Three Sigma security audit.

### DIFF
--- a/packages/protocol/src/LiquidationManager.sol
+++ b/packages/protocol/src/LiquidationManager.sol
@@ -108,13 +108,15 @@ contract LiquidationManager is ILiquidationManager, SystemAccessControl {
    * @dev Checks this address is the flashloan originator. This check applies to a
    * {BorrowingVault} only.
    *
-   * @param vault that holds user's position
    * @param users to be liquidated
+   * @param liqCloseFactors (optional) for each user, otherwise pass zero for each
+   * @param vault that holds user's position
    * @param debtAmount amount to liquidate
    * @param flasher contract address
    */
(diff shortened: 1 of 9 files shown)


DaigaroCota:
Hey @brozorec @pedrovalido I would prefer that both be aware and review this pr 🙏 .
To coordinate workflow, just submit approval, then I will merge after both approvals are complete.

pedrovalido (approved):
Great work! Will wait for @brozorec 's review to merge

packages/protocol/src/LiquidationManager.sol
```diff
@@ -156,17 +180,20 @@ contract LiquidationManager is ILiquidationManager, SystemAccessControl {
    * @notice Callback function that completes execution logic of a liquidation
    * operation with a flashloan.
    *
-   * @param vault that holds user's position
    * @param users to be liquidated
+   * @param liqCloseFactors (optional) for each user, otherwise pass zero for each
```
brozorec, line 184: @DaigaroCota Can you better describe the `liqCloseFactor` in the comment and how it would affect calling this method with zeros or other values?

brozorec: I see actually it in the `IVault` but I guess giving an example could further help the caller in determining this value.

DaigaroCota: Added an example as a dev doc on how to call `{LiquidationManager.liquidate(...)` with args using ethersJs.

PR #638, opened 2023-06-28
XReceive fails to record msg if no gas

DaigaroCota:
This pull request addresses Item #-21 of the Three Sigma security audit.


DaigaroCota:
@0x73696d616f this pull request is still in draft, but please check the comment in attached issue.

DaigaroCota:
@0x73696d616f I believe I finally understood, please review. Also I found that the catch-block when recording a failed txn can vary with our current ConnextHandler-recordFailed(..) implementation.

PR #639, opened 2023-06-28
Introduce xreceiveproxy for safer connextrouter

DaigaroCota:
This pull request addresses Item #-16 of the Three Sigma security audit.

### DIFF
--- a/packages/protocol/src/routers/ConnextRouter.sol
+++ b/packages/protocol/src/routers/ConnextRouter.sol
@@ -12,6 +12,7 @@ pragma solidity 0.8.15;
 import {SafeERC20} from "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import {IERC20} from "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
 import {IConnext, IXReceiver} from "../interfaces/connext/IConnext.sol";
+import {XReceiveProxy} from "./XReceiveProxy.sol";
 import {ConnextHandler} from "./ConnextHandler.sol";
 import {BaseRouter} from "../abstracts/BaseRouter.sol";
 import {IWETH9} from "../abstracts/WETH9.sol";
@@ -23,6 +24,8 @@ import {IFlasher} from "../interfaces/IFlasher.sol";
 import {LibBytes} from "../libraries/LibBytes.sol";
 
 contract ConnextRouter is BaseRouter, IXReceiver {
+  using SafeERC20 for IERC20;
+
   /**
    * @dev Emitted when a new destination router gets added.
    *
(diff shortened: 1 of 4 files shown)


DaigaroCota:
Hey @brozorec , this pull request has a pattern refactor on how the xreceive function will work. 
I drafted the PR so that the changes are evident. 
I still have to update tests and such, however, I wanted to have your opinion on the effects of this changes.
I certainly now this will affect SDK, as the `to_` address in the xReceive call will change.
Please review also the attached issue and the discussion there.

packages/protocol/src/routers/XReceiveProxy.sol
```diff
@@ -0,0 +1,59 @@
...
+    bytes memory callData
+  )
+    external
+    returns (bytes memory)
+  {
+    IERC20 asset_ = IERC20(asset);
+
+    if (amount == 0) {
+      revert XReceiveProxy__xReceive_noValueTransfer();
+    }
```
brozorec, line 49: Why passing calldata only (without transfer of funds) wouldn't be a valid case?

packages/protocol/src/routers/ConnextRouter.sol
```diff
@@ -137,31 +148,25 @@ contract ConnextRouter is BaseRouter, IXReceiver {
...
-    uint256 beforeSlipped;
-    if (amount > 0) {
-      // Ensure that at this entry point expected `asset` `amount` is received.
-      balance = IERC20(asset).balanceOf(address(this));
-      if (balance < amount) {
-        revert ConnextRouter__xReceive_notReceivedAssetBalance();
-      } else {
-        _tempTokenToCheck = Snapshot(asset, balance - amount);
-      }
+    IERC20 asset_ = IERC20(asset);
```
brozorec, line 156: maybe no need to declare an additional variable here, use directly `IERC20(asset)` in the next 2 lines

packages/protocol/src/routers/ConnextRouter.sol
```diff
@@ -82,6 +84,7 @@ contract ConnextRouter is BaseRouter, IXReceiver {
   IConnext public immutable connext;
 
   ConnextHandler public immutable handler;
+  address public immutable xreceiveProxy;
```
brozorec, line 87: xReceiveProxy

brozorec (commented):
That will require some changes in the SDK to replace the param `receiver` from ConnextRouter to the new `XReceiveProxy`. Est. time for that change: no more than 1d

packages/protocol/src/routers/XReceiveProxy.sol
DaigaroCota: You are right, Updating it.

0x73696d616f (approved):
Looks good

brozorec (approved):

PR #640, opened 2023-06-29
Handling borrowingvault external full debt repay

DaigaroCota:
This pull request addresses Item #-43 of the Three Sigma security audit.

### DIFF
--- a/packages/protocol/src/vaults/borrowing/BorrowingVault.sol
+++ b/packages/protocol/src/vaults/borrowing/BorrowingVault.sol
@@ -516,7 +516,7 @@ contract BorrowingVault is BaseVault {
     uint256 debt = convertToDebt(debtShares);
 
     // uint256 baseUserMaxBorrow = ((assets * maxLtv * price) / (1e18 * 10 ** decimals()));
-    uint256 baseUserMaxBorrow = assets.mulDiv(maxLtv.mulDiv(price, 1e18), 10 ** decimals());
+    uint256 baseUserMaxBorrow = assets.mulDiv(maxLtv * price, 10 ** decimals() * 1e18);
     max = baseUserMaxBorrow > debt ? baseUserMaxBorrow - debt : 0;
   }
 
--- a/packages/protocol/src/vaults/borrowing/BorrowingVault.sol
+++ b/packages/protocol/src/vaults/borrowing/BorrowingVault.sol
@@ -29,8 +29,6 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 import {BaseVault} from "../../abstracts/BaseVault.sol";
 import {VaultPermissions} from "../VaultPermissions.sol";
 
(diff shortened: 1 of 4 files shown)


0x73696d616f (approved):
Looks good. I added some comments below, please check. Additionally, maybe the `mulDiv` could be used in `lockedAssets`, `_computeFreeAssets`.

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -29,6 +29,8 @@ import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 import {BaseVault} from "../../abstracts/BaseVault.sol";
 import {VaultPermissions} from "../VaultPermissions.sol";
 
+import {console} from "forge-std/console.sol";
```
0x73696d616f, line 32: console.log is here, probably better to remove before merging

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -544,7 +515,8 @@ contract BorrowingVault is BaseVault {
     uint256 debtShares = _debtShares[borrower];
     uint256 debt = convertToDebt(debtShares);
 
-    uint256 baseUserMaxBorrow = ((assets * maxLtv * price) / (1e18 * 10 ** decimals()));
+    // uint256 baseUserMaxBorrow = ((assets * maxLtv * price) / (1e18 * 10 ** decimals()));
+    uint256 baseUserMaxBorrow = assets.mulDiv(maxLtv.mulDiv(price, 1e18), 10 ** decimals());
```
0x73696d616f, line 519: I think the math is a bit different now, there could be precision loss. This is the new formula, based on the new code
```solidity
assets*(maxLtv*price/1e18)/10**decimals()
```
doing instead
```solidity
assets.mulDiv(maxLtv*price, 10 ** decimals()*1e18);
```
would prevent precision loss
the risk would be `maxLtv*price` overflow

packages/protocol/src/vaults/borrowing/BorrowingVault2.sol
```diff
@@ -423,7 +392,8 @@ contract BorrowingVault2 is BaseVault {
     uint256 debtShares = _debtShares[borrower];
     uint256 debt = convertToDebt(debtShares);
 
-    uint256 baseUserMaxBorrow = ((assets * maxLtv * price) / (1e18 * 10 ** decimals()));
+    // uint256 baseUserMaxBorrow = ((assets * maxLtv * price) / (1e18 * 10 ** decimals()));
+    uint256 baseUserMaxBorrow = assets.mulDiv(maxLtv.mulDiv(price, 1e18), 10 ** decimals());
```
0x73696d616f, line 396: same comment as above

pedrovalido (approved):

PR #644, opened 2023-06-30
Harvesting basic structure

pedrovalido:


packages/protocol/src/interfaces/ILendingProvider.sol
```diff
@@ -124,4 +124,18 @@ interface ILendingProvider {
...
+   * @param data bytes to be used to call the harvest function at the lending provider.
+   *
+   */
+  function harvest(bytes memory data) external returns (bool success);
+
+  /**
+   * @notice Returns the address of the `token` that is harvested from the protocol.
+   *
+   */
+  function getHarvestToken() external view returns (address token);
```
brozorec, line 140: What about there is more than 1 token to harvest, like in the case of Morpho?

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -970,4 +972,48 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+    }
+
+    //strategy 2 = repay debt
+    /// @dev only implemented in BorrowingVault
+    if (strategy == Strategy.RepayDebt) {
+      revert BaseVault__harvest_strategyNotImplemented();
+    }
+
+    //strategy 3 = distribute rewards
+    if (strategy == Strategy.Distribute) {}
```
brozorec, line 1010: Wouldn't it be more flexible if this logic is in the provider?

packages/protocol/src/interfaces/ILendingProvider.sol
pedrovalido: good point. I will change to return address[] tokens . I've also added IVault as parameter because I've noticed a lot of them depend on the asset (makes sense).

packages/protocol/src/abstracts/BaseVault.sol
pedrovalido: Yes, it would. I actually was thinking about changing it because compoundV3 has a claimTo function which can be leveraged if we know the strategy. Well spotted !

pedrovalido: While implementing this, I realised it's not as useful as I thought to have this logic in the provider as it is in the vault. The claimTo function of CompoundV3 cannot be leveraged (we alway need to claimTo(address(vault)):
strategy 1: claim to vault to swap for collateral
strategy 2: claim to vault to swap for debt
strategy 3: cannot claim to all users to distribute rewards

In the general functionality of the harvest:
If we want to swap for collateral or debt (strategy 1 and 2) we will harvest and then swap. The strategy will still be in the ILendingProvider.harvest (and ILendingProvider.previewHarvest) in case we're able to choose which token to get rewards in. In those cases, the preview harvest will return the address[] tokens and uint256[] amounts according to the strategy (ex. previewHarvest will return the collateralAsset and amount if the strategy is convertToCollateral).

The logic present in the vault will be the swap (depending on the strategy and values returned by previewHarvest). If we put the swap logic inside the provider, we will be repeating a lot of code. 

Will do this changes in this PR.

Let me know your thoughts on this.

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -986,4 +995,75 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+  )
+    internal
+    returns (address[] memory tokens, uint256[] memory amounts)
+  {
+    bytes memory callData = abi.encodeWithSelector(provider.harvest.selector, data);
+    (bytes memory returnData) = address(provider).functionDelegateCall(
+      callData, string(abi.encodePacked("harvest", ": delegate call failed"))
+    );
+    (tokens, amounts) = abi.decode(returnData, (address[], uint256[]));
+  }
```
DaigaroCota, line 1056: There should be some type of event emitted here, essentially for all the tokens and amounts harvested and the provider.

Same applies for the upgradeable vault.

packages/protocol/src/HarvestManager.sol
```diff
@@ -0,0 +1,270 @@
...
+    }
+    if (strategy == Strategy.RepayDebt) {
+      data = _repayDebt(provider, tokens, amounts, swapper, vault_);
+    }
+    if (strategy == Strategy.Distribute) {
+      data = _distribute(tokens, amounts, vault_);
+    }
+
+    currentVaultHarvest = address(0);
+  }
```
DaigaroCota, line 143: Completing the harvest through the HarvestManager should also emit an Event. 
This should not overlap with the information on the event that should also be emitted on the vault itself. 
I probably suggest that this event here focuses on the treasury profit and what the vault "gained" by the strategy.

packages/protocol/src/swappers/UniswapV2Swapper.sol
```diff
@@ -109,7 +109,7 @@ contract UniswapV2Swapper is ISwapper {
   {
     address[] memory path = _buildPath(assetIn, assetOut);
     uint256[] memory amounts = uniswapRouter.getAmountsOut(amountIn, path);
-    amountOut = amounts[1];
+    amountOut = amounts[amounts.length - 1];
```
DaigaroCota, line 112: good catch here 👍

packages/protocol/src/vaults/borrowing/BorrowingVault.sol
```diff
@@ -24,15 +24,19 @@ import {
...
 import {SafeERC20} from "openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol";
 import {Math} from "openzeppelin-contracts/contracts/utils/math/Math.sol";
 import {BaseVault} from "../../abstracts/BaseVault.sol";
 import {VaultPermissions} from "../VaultPermissions.sol";
+import {ISwapper} from "../../interfaces/ISwapper.sol";
 
 contract BorrowingVault is BaseVault {
   using Math for uint256;
   using SafeERC20 for IERC20Metadata;
+  using SafeERC20 for IERC20;
```
DaigaroCota, line 39: Since IERC20Metadata is IERC20, just feel free to use IERC20Metadata. Same applies to upgradeable borrowing vault.

DaigaroCota (changes requested):

packages/protocol/src/HarvestManager.sol
```diff
@@ -0,0 +1,270 @@
...
+      (strategy == Strategy.ConvertToCollateral || strategy == Strategy.RepayDebt)
+        && !chief.allowedSwapper(address(swapper))
+    ) {
+      revert HarvestManager__harvest_notValidSwapper();
+    }
+
+    if (currentVaultHarvest != address(0)) {
+      revert HarvestManager__harvest_harvestAlreadyInProgress();
+    }
+
```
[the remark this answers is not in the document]
pedrovalido: Added a verification using the vault status (chief.isVaultActive). what do you think?

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -986,4 +995,75 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+    ISwapper swapper,
+    bytes memory data
+  )
+    external
+    virtual
+    hasRole(msg.sender, HARVESTER_ROLE)
+    returns (address[] memory tokens, uint256[] memory amounts)
+  {
+    //check strategy is valid for this vault
+    if (strategy == Strategy.RepayDebt) {
```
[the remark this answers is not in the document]
pedrovalido: Good idea. I will add a TODO comment to be aware of this when pushing down the contract size

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -986,4 +995,75 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+    }
+    //check provider is valid
+    if (!_isValidProvider(address(provider))) {
+      revert BaseVault__harvest_invalidProvider();
+    }
+
+    //collect rewards from provider
+    (tokens, amounts) = _harvest(provider, data);
+
+    if (tokens.length != amounts.length) {
```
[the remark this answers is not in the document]
pedrovalido: Makes sense :) Will add a TODO comment here as well

packages/protocol/src/swappers/UniswapV2Swapper.sol
pedrovalido: thank you 🤓

PR #647, opened 2023-07-03
BaseVault ensure ERC compliance

DaigaroCota:
This pull request addresses Item #-37 of the Three Sigma security audit.


0x73696d616f (approved):
I think the approval only letting the spender be the receiver could surprise other devs, but other than that it's good

pedrovalido (approved):

PR #651, opened 2023-07-04
Router decoding bug and revert when receiver is router

DaigaroCota:
This pull request addresses Item #-25 and #-26 of the Three Sigma security audit.

### DIFF
--- a/packages/protocol/src/routers/ConnextRouter.sol
+++ b/packages/protocol/src/routers/ConnextRouter.sol
@@ -329,8 +329,7 @@ contract ConnextRouter is BaseRouter, IXReceiver {
       bytes memory callData
     ) = abi.decode(params, (uint256, uint256, address, uint256, address, bytes));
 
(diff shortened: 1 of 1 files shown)


0x73696d616f (changes requested):
Looks good, but there is still a decoding bug in [`_getBeneficiaryFromCalldata()`](https://github.com/Fujicracy/fuji-v2/blob/main/packages/protocol/src/routers/ConnextRouter.sol#L415)

DaigaroCota:
> Looks good, but there is still a decoding bug in [`_getBeneficiaryFromCalldata()`](https://github.com/Fujicracy/fuji-v2/blob/main/packages/protocol/src/routers/ConnextRouter.sol#L415)

eagle eye 🦅 , thanks for noting 🙌

0x73696d616f (approved):
Looks good

pedrovalido (approved):

PR #657, opened 2023-07-11
Connexthandler set executed storage true-false directly

DaigaroCota:
This pull request addresses Item #-24 of the Three Sigma security audit.


0x73696d616f (approved):
Looks good

PR #660, opened 2023-07-12
Deployment fixes borrowing vault and factory

DaigaroCota:
I identified in the following [tx](https://optimistic.etherscan.io/tx/0xb02949c668f17bc8a3984aec206d51615420e5ccc47923229997e40b07be9684) that paying more than max leaves unattended assets in the vault. 

Therefore the fix is required.  This is an extension of what we discussed that must be considered in the `withdraw` if we want to have the same default to max behavior.


DaigaroCota:
@brozorec I will also add for `withdraw` as we were discussing earlier today.

brozorec (approved):

PR #661, opened 2023-07-12
Withdraw more than max replaces for max

DaigaroCota:
- [x]  The withdraw function can handle withdraw max softly instead of revert. Applies to redeem too.
- [x]  Took a chance to optimize how similar payback max is handled. 
- [x] Added the discussed file for that prints the errors 4byte selector.
- [x] Updated and added test accordingly.


packages/protocol/src/vaults/borrowing/BorrowingVaultUpgradeable.sol
```diff
@@ -548,6 +526,45 @@ contract BorrowingVaultUpgradeable is BaseVaultUpgradeable {
...
+    _payback(caller, owner, debt_, shares_);
+
+    if (remainder > 0) {
+      /**
+       * @devSince the `_payback(...) only pulls (erc20) that is needed to payback
+       * maxAmount, this logic handles excess amount `remainder` by pulling from
+       * `msg.sender and returning to the `owner` the `remainder`.
+       */
+      _debtAsset.safeTransferFrom(caller, address(this), remainder);
+      _debtAsset.safeTransfer(owner, remainder);
```
brozorec, line 564: Why not do this with one call?
`_debtAsset.safeTransferFrom(caller, owner, remainder);`

A similar note for "withdraw".

packages/protocol/src/abstracts/BaseVault.sol
```diff
@@ -550,6 +543,39 @@ abstract contract BaseVault is ERC20, SystemAccessControl, PausableVault, VaultP
...
+    address caller,
+    address receiver,
+    address owner
+  )
+    internal
+    returns (uint256 assets_, uint256 shares_)
+  {
+    // This local var helps save gas not having to call provider balances again.
+    // and it is multiplied by asset decimals to mantain precision.
+    uint256 savedAssetSharesRatio = shares > 0 ? assets.mulDiv(10 ** (decimals()), shares) : 0;
```
brozorec, line 568: why does this need to be assigned here? couldn't be moved to `_withdrawChecks`?

DaigaroCota:
@brozorec implemented suggested changes.

brozorec (approved):

PR #678, opened 2023-07-19
Referral program

ferostabio:


brozorec:
Great, it's already working :ok_hand: :fireworks: Spotted some issues while testing (maybe we need to update from main to get some of them fixed?):
1. Paddings in "Position Overview" on My position page 
image
2. Form in My positions page
image
3. Referral modal: can we put all of those elements on a single line
image
4. How can we change the wording here
image
and here
image

NikolaiYurchenko:
Fixed some of comments, for others I beleive we need to dig into admin panel of a referral program

ferostabio:
> Fixed some of comments, for others I believe we need to dig into admin panel of a referral program

Point 3. I reported last Friday, Jean-Louis took a screenshot of it. Might be a good idea to report it again @brozorec .

Point 4... I'm in the admin panel and it seems I can't modify anything, we should probably ask them to do it for us.

PR #680, opened 2023-07-19
Simplified navigation bar

### LINKED ISSUE #670: Simplified navigation bar

### LINKED ISSUE #671: Navigation Bar - Positions at Risk

### LINKED ISSUE #683: Account modal

ferostabio:
close #670 
close #671 
close #683

- [x] Calculate positions at risk 
- [x] New simplified bar
- [x] Show positions at risk
- [x] Show healthy positions and no positions
- [x] Calculate all balances
- [x] Update account modal
- [x] Account modal's recent transactions
- [x] Account modal's asset list
- [x] add loader
- [x] fix long asset nage styles
- [x] fix long collateral name in markets screen (look at staked ETH)

Captura de pantalla 2023-07-31 a la(s) 15 58 29

- [x] when a user reaches either the borrowing screen or the lending screen, we show previously selected vaults before flashing and changing to the right ones
- [x] when a user reaches the screens, before changing a chain, most of the times we don't show rewards (meaning financials aren't loaded)

### DIFF
--- a/packages/frontend/components/Shared/TabSwitch/Chip.tsx
+++ b/packages/frontend/components/Shared/TabSwitch/Chip.tsx
@@ -33,7 +33,7 @@ function Chip({ selected, label, onClick, size }: TabChipProps) {
           variant="small"
           color={palette.text.primary}
           fontSize={size === 'large' ? '1rem' : '0.875rem'}
-          lineHeight="100%"
+          lineHeight="120%"
           textAlign="center"
           sx={{
             ['@media screen and (max-width: 980px)']: {
--- a/packages/frontend/components/Shared/Table/IntegratedProviders.tsx
+++ b/packages/frontend/components/Shared/Table/IntegratedProviders.tsx
@@ -1,8 +1,9 @@
-import { Box, Chip, Stack, Tooltip } from '@mui/material';
+import { Box, Chip, Stack } from '@mui/material';
 import React from 'react';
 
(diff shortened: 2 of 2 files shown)


ferostabio:
@brozorec have a look! we're showing positions at risk based on ltvMax - ltv, what would be the right amount to consider a position at risk? or would you like a different approach?

brozorec:
Good job :muscle: I like it, like also the idea to have it as a loader.

> @brozorec have a look! we're showing positions at risk based on ltvMax - ltv, what would be the right amount to consider a position at risk? or would you like a different approach?

I'd say we consider a position at risk when ltvMax - ltv < 10%

---

2 suggestions:
1. when the notification circle is green and the user clicks on it, we open the notification popover with a message instead of a tooltip as it's proposed in the design. This will give us some space to put a CTA "Create notification" in the future when we are going to implement this feature.
2. When the user hovers over the network icon, we unfold the name with the dropdown instead of displaying a tooltip.

wdyt @Markoyw ?

NikolaiYurchenko:
In case of success state add menu content  to show 
1st case ( no positions )
2nd case ( position are healthy )

ferostabio:
@NikolaiYurchenko the first time we load the new account modal, there's a slight delay before we show assets. To be expected, of course. Lets add the same type of skeleton we show for markets, since having an empty list looks a bit weird. Show enough elements so they cover all the screen, since even before adding more chains we already have more than 10 tokens.

PR #681, opened 2023-07-20
Router replace amout after soft max withdrawal

DaigaroCota:
This pull request addresses the router issue of replacing arguments when the encoded amounnt-arg in actions after a Withraw happens differs from the "real" amount received.


DaigaroCota:
Hey @brozorec  ready for review: 
Issue related to:
```
The other issue is in the router. After doing the a soft-withdrawal, I want to cross-transfer but as I withdrew less than what was encoded in the cross-transfer action
```
The PR actually addresses all the Action cases after a soft-max withdraw.

DaigaroCota:
I checked the refactors, they all look good on my side.

brozorec (approved):

PR #693, opened 2023-08-02
Restrict amounts when bridging to Ethereum

### LINKED ISSUE #697: Hot-fix restricting to-eth amounts
Many users have experience problems transfering small amounts to ethereum.

We should restrict users for sending "silly" amounts to ethereum. 

Threshold minimum 1000 usd value.

ferostabio:
close #697


ferostabio:
Per the #development conversation, I started restricting the amounts to be transacted. If the user is attempting to bridge less than a thousand dollars to Ethereum, we'll disable the button. A couple of notes:

- So far I implemented it in lending, since its simpler.
- Both the min amount and shown text are constants so changing the values should be as straightforward as it gets.
- Adding this temporary feature to borrowing is a matter of: 
  -  there could be three chains, so we need to find if theres a bridging operation to Ethereum
  -  and the bridge might involve either the collateral or the debt

ferostabio:
It should be done now

packages/frontend/helpers/assets.ts
```diff
@@ -291,3 +295,33 @@ export const withdrawMaxAmount = async (
...
+
+export const invalidBridgingAmount = (
+  steps: RoutingStepDetails[],
+  collateral: AssetChange,
+  debt?: AssetChange
+) => {
+  for (const step of steps) {
+    if (step.step === RoutingStep.X_TRANSFER && step.token) {
+      const asset = [collateral, debt]
+        .filter((asset) => asset)
```
brozorec, line 307: what's the point of this filter

packages/frontend/components/Shared/Operation/OperationButton.tsx
```diff
@@ -20,6 +20,7 @@ export enum OperationButtonTitles {
   SIGN = 'Sign & ',
   WITHDRAW = 'Withdraw',
   WITHDRAW_MAX = 'Withdraw more than allowed',
+  ETHEREUM_MIN = 'Bridging to Ethereum minimum of $1000',
```
brozorec, line 23: Min. $1000 required to bridge to Ethereum

brozorec (approved):
good job :+1: added some minor points

packages/frontend/helpers/assets.ts
ferostabio: First, we want only `X_TRANSFER` steps since the `steps` array contains all of them. Then we make sure it has a token since its optional and finall grab the asset since it already contains price in USD

PR #695, opened 2023-08-03
Disable wstETH bridging

### LINKED ISSUE #696: Handle Unbridgeable token

ferostabio:
close #696 

Added `wstETH` to a no-bridging list. If, in a crosschain operation, its required `wstETH` to be bridged, we're no longer allowing that.


brozorec:
It's better to inverse the logic and checking if the token is bridgeable is better. We'll start adding more tokens that are not bridgable (cbETH, rETH, GNO, ...), and the chance of miss to add the token to the list is higher. In the current setup, Connext will always be slower in adding more tokens because they also need to bootstrap the liquidity for them.

PR #699, opened 2023-08-04
Aave v3 efficiency mode

DaigaroCota:
This pull requests adds the efficiency mode to the Aave V3 lending providers.


DaigaroCota:
@brozorec this PR is 99% ready, only missing step is to actually redeploy the AaveV3 lending provider in all chains. 
All forking tests with AaveV3 are passing, therefore I assume it should be fine now on real nets. 
I followed your suggestion and now the emode configs are part of each chain json.
Please review and if you have some extra time we can do a test for replacing the provider on a L2. Once we are good there, then we move to do it in all chains.

DaigaroCota:
@brozorec all new vaults that require e-mode have been updated and check for functionality.

These vaults use AaveV3 provider address with the emode upgrade. All other vaults can remain using old AaveV3, since e-mode is not applicable to their collateral-debt asset pair.

arbitrum:

- BorrowingVault-RETHWETH-1 0x9E41477315c4045B129351428147761082762AEE
- BorrowingVault-WSETHWETH-1 0x4DB1d229170417eDE991E20BE159C0847D67c9aD
- BorrowingVault-WSETHWETH-2 0x30E214b05CdA6438ad35b2597130C338cA8AEF74

optimism:
- BorrowingVault-RETHWETH-1 0xEaBDA502948192D8dF49AB91Fc278ec9C8C65ef8
- BorrowingVault-WSTETHWETH-1 0x389f1739053D2273bAe38CBb878b938AF4cDfC85

ethereum
- BorrowingVault-WSTETHWETH-1 0xBf2b58B962cB951FCEa3F32e3afd54661aA2A78e
- BorrowingVault-CBETHWETH-1 0x288465D0385aed5971467E84718BDCE553e24498
- BorrowingVault-RETHWETH-1 0xb2a57D48a1553B071B99353CC07dCc1a2D2730c9

brozorec (changes requested):
1. You have overridden the file with the address of aave v3 everywhere. This will create issues bc we loose track of the address for this provider for some of the vaults that are still using it. If we are going to use both contracts, then better keep them both: for ex: `Aave_V3_Arbitrum` and `Aave_V3_Arbitrum_Emode`. This will facilitate handling both providers on the frontend as well bc rn it messes them up.
2. Why some of the emode configs are different for the same pairs on different chains? (I guess this should be ok, but just curious)
- Arbitrum:
```
    {
      "asset" : "WSTETH",
      "debtAsset": "WETH",
      "id": 2
    }
```
- Ethereum:
```
    {
      "asset" : "WSTETH",
      "debtAsset": "WETH",
      "id": 1
    }
```

DaigaroCota:
> 1. You have overridden the file with the address of aave v3 everywhere. This will create issues bc we loose track of the address for this provider for some of the vaults that are still using it. If we are going to use both contracts, then better keep them both: for ex: `Aave_V3_Arbitrum` and `Aave_V3_Arbitrum_Emode`. This will facilitate handling both providers on the frontend as well bc rn it messes them up.

Agree to keep them with different names. 

> 2. Emode config numbers are set by Aave governance. On the question on why they decided to not keep them consistent across chains, I think is up to them 😕 . From what I observed on other chains they set first the stablecoin-correlated configs and then ethereum-correlated. However, on the AaveV3-ethereum they set first the ethereum-correlated configs, and they haven't set-up any stablecoin-correlated config. I guess they are set as they go. For instance on Polygon the matic-correlated is configId=3, but I don't think this will be applicable to any other chain.

PR #707, opened 2023-08-10
Spark lending provider

pedrovalido:
Integration of spark as lending provider. 

Note: 
- Had to do a small change in the test. Usdc borrowing is not enabled yet so just used dai instead.


DaigaroCota (changes requested):
@pedrovalido lest consider emode for Spark. Please update this branch with that of #699.
This will be helpful for higher LTV vaults 💪

DaigaroCota (approved):
This PR is approved, but not merged yet. 
Since its code changes are branching off  from #699 , we shall wait for that PR to merge to main, and then this will follow.

PR #720, opened 2023-08-16
Frontend cache for llamas

### LINKED ISSUE #717: Defillama cache
We experience `403` from Defillama getting data. We are already [working](https://github.com/orgs/Fujicracy/projects/2/views/12?pane=issue&itemId=35686620) on a custom API to cache the data, but the underlying issue is that their response weights 8mb. If we just move this to our own server and we make a lot of calls, we'll have issues maintaining it.

Also, its bad user experience. We are making a lot of calls, not just for the markets screen, also for the charts, downloading 8mb of data each time.

We are then going to cache the data locally with `indexedDB` and make the calls to Defillama (first, then to our own server) each X amount of time.

ferostabio:
close #717 

In order to reduce the number of calls we make to an API (Defillama or ours) and before deploying our own express + redis API, I added a local cache with `indexedDB`.

Now we will only call an API to get the data once in fifteen minutes -rest of the time we'll load data locally. 

In order to do this, I had to decouple to functions on the sdk that were calling an API and parsing through its response.

### DIFF
--- a/packages/frontend/helpers/cache.ts
+++ b/packages/frontend/helpers/cache.ts
@@ -66,13 +66,16 @@ export async function getLlamaCache(): Promise<GetLLamaFinancialsResponse> {
   });
 }
 
-export async function getCacheLastUpdatedDate(): Promise<Date> {
-  return new Promise(async (resolve, reject) => {
+export async function getCacheLastUpdatedDate(): Promise<Date | undefined> {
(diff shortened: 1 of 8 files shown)


ferostabio:
@DaigaroCota @NikolaiYurchenko please have a look 🚀

brozorec (approved):

PR #721, opened 2023-08-17
UI issues

NikolaiYurchenko:


ferostabio:
@NikolaiYurchenko I was playing around with the work you did, and I think everything we're doing is actually worst. Now... I realized that the underlying issue is that we have one `MarketsTable` with two different data sources. Let's just have two tables and show one of them according to the selected tab! No transition will be needed (`tableData` will have to be an array containing both sets of data)

ferostabio:
Since you're at it, please have a look at this tiny thing, "..." is not centered.

Captura de pantalla 2023-08-17 a la(s) 11 17 40

ferostabio:
Thanks for the fixes @NikolaiYurchenko 😁

The glitch is still there, though. Wasn't expecting that, since they are now different tables. I would say...

1 - Try to see if there is a way out of it. They are different tables after all now.
2 - If you find no way out, do implement the loader... but not just one row. We have the row count for each list, so we can now how many skeleton rows to show (and for a bit less time)

ferostabio:
Another small UI issue @NikolaiYurchenko ! Left panel shouldn't change width when the "...after..." component appears.

https://github.com/Fujicracy/fuji-v2/assets/501915/878295e0-25c8-48c2-861d-49a144af1cec

NikolaiYurchenko:
> Thanks for the fixes @NikolaiYurchenko 😁
> 
> The glitch is still there, though. Wasn't expecting that, since they are now different tables. I would say...
> 
> 1 - Try to see if there is a way out of it. They are different tables after all now. 2 - If you find no way out, do implement the loader... but not just one row. We have the row count for each list, so we can now how many skeleton rows to show (and for a bit less time)

well, it will be blinking just because it is insta-changing, moreover on a dark background

ferostabio:
> well, it will be blinking just because it is insta-changing, moreover on a dark background

If you play the video in slow motion the issue is super clear 😅 (do vlc at .25x). For a brief moment, while switching tabs, you get lending data on the borrowing table and borrowing data on the lending one. 

Have a look at the attached screenshot. Have I switched from borrowing to lending, and while its going to show the right info, you can see here that its showing borrowing data. Since we move this data around, its really visible and it produces the glitch.

1 - Switched to lending, showing borrowing data

Captura de pantalla 2023-08-17 a la(s) 15 04 36

2 - Finally shows lending data

Captura de pantalla 2023-08-17 a la(s) 15 06 21

3 - Same thing happens going back to borrowing

Markoyw:
Agree, I like it a lot!

PR #722, opened 2023-08-17
Rebalancer manager error with zero debt rebalance

DaigaroCota:
I noticed that in rebalancing borrowing vaults that only had assets (no debt) the flashloan is not needed. There was a case in where flashloaning zero debt resulted in a "successful" but useless tx that got the Rebalancer locked because it wrote to the `entryPoint` state but never completed the execution and therefore the Rebalancer is locked from being used again.

The fixes of this PR avoid using flashloan when rebalancing a borrowing vault that only has assets ( or zero debt), and a new function restricted to the timelock is added to unblock the `entryPoint` variable in case is necessary. The reason why this is restricted to the timelock is to avoid, reentrancies by malicious `allowedExecutor` users who could figure out ways to by-pass checks of the `entryPoint`.

### DIFF
--- a/packages/protocol/deploy-configs/ethereum.json
+++ b/packages/protocol/deploy-configs/ethereum.json
@@ -198,6 +198,15 @@
       "rating": 85
     },
     {
+      "collateral": "WETH",
+      "debt": "DAI",
+      "liqRatio": 820000000000000000,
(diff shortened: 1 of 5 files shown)


pedrovalido (approved):
LGTM, well spotted :)

